PR #449, opened 2026-05-23
fix(setup,paths): stop silently clobbering host config under transient AKM_STASH_DIR

itlackey:
## Summary

`akm setup --yes --dir .` was silently rewriting the user's host `~/.config/akm/config.json` even when `AKM_STASH_DIR` and `AKM_DATA_DIR` were set to point at a temp directory. The env isolation correctly redirected stash and data file paths but did NOT redirect config persistence. When the temp stash was later cleaned up, the user's global config was left pointing at a deleted path; akm silently fell back to `~/akm` for stash operations so the breakage was invisible until they explicitly inspected the config.

The bug is documented in [`docs/technical/incidents/2026-05-23-setup-clobbers-user-config.md`](./docs/technical/incidents/2026-05-23-setup-clobbers-user-config.md) (now committed) — including reproduction, forensic trail, root cause, and recovery procedure for anyone hit by it before this lands.

This is the same incident that motivated the release-readiness audit's verification tripwire, which itself silently corrupted my own host config for ~5.5 hours.

### Layer 1: `assertSetupSandbox` (in `src/setup/setup.ts`)

Refuses
```
akm setup --dir <transient/path>
```
under all runtimes — not just `bun test` — unless `AKM_FORCE_SETUP_TMP_STASH=1` is set. Wired into all three entry points: `runSetupWizard`, `runSetupWithDefaults`, `runSetupFromConfig`. Mirrors the existing `assertInitSandbox` in `src/commands/init.ts` but fires outside the test runner too. New `SETUP_TMP_STASH_REFUSED` `ConfigError` code with a canned hint pointing at the escape hatch.

### Layer 2: `getConfigDir` / `getCacheDir` isolation rule (in `src/core/paths.ts`)

When `AKM_STASH_DIR` points at a transient path family (`/tmp/`, `/var/tmp/`, `/private/var/folders/`) AND no explicit config/cache override is set (`AKM_CONFIG_DIR`, `XDG_CONFIG_HOME`, `AKM_CACHE_DIR`, `XDG_CACHE_HOME`), config and cache writes route into `${AKM_STASH_DIR}/.akm/` and `${AKM_STASH_DIR}/.akm/cache/` respectively.

Even when the operator opts in to `AKM_FORCE_SETUP_TMP_STASH=1`, the host config remains untouched — config writes follow the stash.

**Crucially:** daily users with a persistent `AKM_STASH_DIR=~/my-stash` see no change. The new rule only fires for transient path families. The precedence ordering means existing tests that pre-arrange `XDG_CONFIG_HOME` for isolation continue to work as before.

### Helper

`isTransientStashPath()` exported from `src/core/paths.ts` and reused by both the path resolvers and the setup guard, so the definition of "transient" stays single-sourced.

## Why both layers?

- Layer 1 fails fast and gives operators a clear error.
- Layer 2 is the structural backstop: if anything else in akm grows a `saveConfig()` call site that gets invoked under an isolated stash, the host config is still protected — no second round of incidents.

## Test plan

- [x] `tests/paths.test.ts` — extends `getConfigDir`/`getCacheDir` coverage for transient `AKM_STASH_DIR`:
  - `AKM_STASH_DIR=/tmp/X` redirects config to `/tmp/X/.akm` and cache to `/tmp/X/.akm/cache`
  - `AKM_STASH_DIR=/var/tmp/X` and `/private/var/folders/...` (macOS mktemp) also redirect
  - `AKM_STASH_DIR=~/persistent` does NOT redirect (daily users unaffected)
  - `AKM_CONFIG_DIR`, `XDG_CONFIG_HOME`, `AKM_CACHE_DIR`, `XDG_CACHE_HOME` all win over the new rule
  - Windows: `APPDATA` also wins
- [x] `tests/setup-tmp-stash-guard.test.ts` — end-to-end coverage:
  - `runSetupWithDefaults({ dir: /tmp/X })` rejects with `SETUP_TMP_STASH_REFUSED` by default
  - `runSetupFromConfig({ dir: /tmp/X })` rejects the same way
  - `AKM_FORCE_SETUP_TMP_STASH=1` lets the call succeed (returns `{ stashDir: /tmp/X }`)
  - Persistent `--dir` (under `~/`) is NOT refused
  - With the escape hatch enabled, a synthesized host `config.json` in a fake HOME is byte-identical (mtime unchanged) after setup runs — and the isolated config landed at `$STASH/.akm/config.json`
- [x] Full suite: **3531 pass / 0 fail / 21 skip** (was 3515 → +16 new tests, zero regressions)
- [x] `bunx tsc --noEmit` → 0 errors
- [x] `bun run lint` → 0 errors (2 pre-existing warnings unchanged)
- [ ] Manual smoke after merge: confirm `akm setup --dir /tmp/x` exits 2 with the migration hint
- [ ] Manual smoke: confirm `AKM_FORCE_SETUP_TMP_STASH=1 akm setup --dir /tmp/x` succeeds and writes to `/tmp/x/.akm/config.json` (host config untouched)

### DIFF
--- a/docs/technical/incidents/2026-05-23-setup-clobbers-user-config.md
+++ b/docs/technical/incidents/2026-05-23-setup-clobbers-user-config.md
@@ -1,0 +1,112 @@
+# Bug: `akm setup --yes --dir .` writes to user's global config even with `AKM_STASH_DIR` set
+
+**Discovered:** 2026-05-23
+**AKM version:** 0.8.0-rc.4 (`release/0.8.0` @ 0e1d339)
+**Severity:** High — silent corruption of user's global config; only detectable by explicit `akm config get stashDir` check.
+
+## Summary
+
+`akm setup --yes --dir .` writes its resulting config to the user's **global** `~/.config/akm/config.json` even when `AKM_STASH_DIR` and `AKM_DATA_DIR` env vars are exported to point at a temp directory. The env isolation correctly redirects stash and data file paths but does **not** redirect config persistence. When a test harness or smoke-test script then removes the temp stash, the user's global config is left with `stashDir` pointing at a deleted path. AKM silently falls back to `~/akm` for stash operations so the breakage is invisible to the user until they explicitly inspect the config.
+
+## Reproduction
+
+```bash
+# Confirm starting state
+akm config get stashDir
+# → "/home/founder3/akm"
+
+# Run a setup smoke test in an isolated temp stash
+cd /tmp && rm -rf repro-stash && mkdir -p repro-stash && cd repro-stash && \
+  AKM_DATA_DIR=$(pwd)/data AKM_STASH_DIR=$(pwd) \
+  bun /path/to/akm/dist/cli.js setup --yes --dir .
+
+# Re-check the global config from the user's shell
+akm config get stashDir
+# → "/tmp/repro-stash"   ← BUG: env-isolated setup leaked into global config
+
+# Worse, after cleanup:
+cd / && rm -rf /tmp/repro-stash
+akm config get stashDir
+# → "/tmp/repro-stash"   ← still pointing at a deleted dir
+akm improve  # silently falls back to ~/akm; user has no idea config is broken
+```
+
+## Real-world impact (incident on this machine)
+
+A parallel Claude Code session (sessionId `e702fb7d-1095-42a7-98eb-e71974457c77`, working in `~/code/github/itlackey/akm/` on the 0.8.0 release-readiness audit) ran:
+
+```bash
+# 06:28:35Z
+cd /tmp && rm -rf tw-stash && mkdir -p tw-stash && cd tw-stash && \
+  AKM_DATA_DIR=$(pwd)/data AKM_STASH_DIR=$(pwd) \
+  /home/founder3/code/github/itlackey/akm/dist/cli.js setup --yes --dir .
+
+# 06:28:48Z
+cd /tmp/tw-stash && rm -rf data .akm; \
+  AKM_DATA_DIR=$(pwd)/data AKM_STASH_DIR=$(pwd) \
+  bun /home/founder3/code/github/itlackey/akm/dist/cli.js setup --yes --dir .
+
+# 06:34:29Z
+rm -rf /tmp/tw-stash && echo "cleaned"
+```
+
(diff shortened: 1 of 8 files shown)


itlackey:
Addressed all review comments + fixed the 3 CI test failures in `9854514`.

## CI failures (3 tests)

Existing tests in `tests/config.test.ts` and `tests/paths.test.ts` deleted `XDG_*_HOME` but did not delete `AKM_STASH_DIR`. Under CI, `AKM_STASH_DIR` was inherited from outer test isolation pointing at a transient dir, which triggered the new isolation rule and overrode the expected HOME-based fallback. Added explicit `delete process.env.AKM_STASH_DIR` to each affected test. Now passes locally.

## Copilot comments

1. **Hint message says config auto-routes but it requires `AKM_STASH_DIR`** (`src/setup/setup.ts:66`, `src/core/errors.ts:78`) — added `applyStashIsolationToEnv()` in setup that pre-sets `process.env.AKM_STASH_DIR = stashDir` when `--dir` is provided (and only when the operator hasn't already set it explicitly). Wired into all 3 setup entry points. Hints reworded to match the actual behavior. New test `--dir /tmp/X without pre-set AKM_STASH_DIR routes config into the stash too` reproduces your exact flagged scenario and asserts the synthesized host config is byte-identical after setup.

2. **`isTransientStashPath` missed `/var/folders/*`** (`src/core/paths.ts:46`) — added the bare check. `os.tmpdir()` on macOS returns `/var/folders/...`; `path.resolve()` does not canonicalize to `/private/var/folders`.

3. **Windows `getCacheDir` precedence comment mismatch** (`src/core/paths.ts:150`) — reordered the function so Windows base dirs (`LOCALAPPDATA` → `USERPROFILE` → `APPDATA`) win before the transient-stash rule, matching `getConfigDir`. Comment now accurate.

4. **4 stale `BUG-setup-clobbers-user-config.md` path references** — all updated to `docs/technical/incidents/2026-05-23-setup-clobbers-user-config.md` (the actual committed location).

## Verification

- Suite: **3533 pass / 0 fail / 21 skip** (was 3531 → +2 new tests, zero regressions)
- `bunx tsc --noEmit`: exit 0
- `bun run lint`: exit 0 (2 pre-existing warnings only)

Thank you for the thorough review!

PR #488, opened 2026-05-25
docs: align install URL with release artifact

### LINKED ISSUE #479: README + getting-started point at two different install URLs
\`docs/getting-started.md:16,19\` uses \`raw.githubusercontent.com/.../install.sh\` URLs while \`README.md:17,20\` uses \`releases/latest/download/...\`. Both work, but they are different code paths — the raw-githubusercontent one installs whatever is on \`main\`, which may be ahead of the latest published release.

Pick one canonical URL and use it everywhere. Recommend \`releases/latest/download/install.sh\` because it is pinned to a tagged artifact (with checksum already verified via \`release.yml:185-188\`).

Files: \`README.md\`, \`docs/getting-started.md\`, plus \`.github/README.npm.md\` if it diverges.

narutamaaurum:
## Summary
- switch the README standalone installer command to the release artifact URL
- keep the install instructions aligned with the tagged-release path described in #479

## Validation
- verified `README.md` now uses `releases/latest/download/install.sh`
- verified `docs/getting-started.md` and `.github/README.npm.md` do not still reference the raw `main` installer URL
- `bun` is not installed in this runner, so no project checks were available for this docs-only change

Closes #479

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Everything else supports one of those steps.
 
 ```sh
 # Standalone binary (no runtime dependencies)
-curl -fsSL https://raw.githubusercontent.com/itlackey/akm/main/install.sh | bash
+curl -fsSL https://github.com/itlackey/akm/releases/latest/download/install.sh | bash
 
 # Or via Bun
 bun install -g akm-cli


itlackey:
Thank you for catching that!

PR #583, opened 2026-06-09
port: 0.8.6 — consolidate.incrementalSince profile config field

itlackey:
Forward-port of the 0.8.6 hotfix to main (0.9.0).

Cherry-pick of `da949c03` from `hotfix/0.8.6`.

## What

Surfaces `incrementalSince` as a configurable field in the `consolidate` block of an improve profile. Setting it (e.g. `"7d"`) narrows the candidate pool to memories modified within that window plus their top-5 graph neighbours — making it practical to run consolidation more than once per day without re-scanning the full pool each time.

The `incrementalSince` option already existed in `akmConsolidate()` but was hardcoded off at the call site after the 0.8.5 full-pool sweep fix. This change makes it opt-in via config rather than hardcoded.

## Files changed

- `src/core/config/config-schema.ts` — add `incrementalSince: z.string().min(1).optional()` to `ImproveProcessConfigSchema`
- `src/commands/improve/improve.ts` — read `improveProfile?.processes?.consolidate?.incrementalSince` and pass to `akmConsolidate()`
- `schemas/akm-config.json` — regenerated
- `CHANGELOG.md` — 0.8.6 entry (note: version number stays at 0.9.0 on main)
- `.github/workflows/release.yml` — allow `hotfix/*` branches to dispatch stable releases

## Testing

`bun test` — 4172 pass, 0 fail

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,8 +40,10 @@ jobs:
             echo "Dispatched from default branch ($BRANCH)."
           elif [[ "$BRANCH" == release/* ]] && echo "$VERSION" | grep -qE '-'; then
             echo "Dispatched from release branch ($BRANCH) with prerelease version ($VERSION)."
+          elif [[ "$BRANCH" == hotfix/* ]] && ! echo "$VERSION" | grep -qE '-'; then
+            echo "Dispatched from hotfix branch ($BRANCH) with stable version ($VERSION)."
           else
-            echo "Releases must be dispatched from $DEFAULT, or from a release/* branch with a prerelease version (e.g. 0.6.0-rc1)."
+            echo "Releases must be dispatched from $DEFAULT, from a release/* branch with a prerelease version, or from a hotfix/* branch with a stable version."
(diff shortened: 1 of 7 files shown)


itlackey:
### Bug fix added: `incrementalSince` duration strings were broken

Discovered during testing: `"30m"`, `"24h"`, `"7d"` config values were being passed raw to `narrowToIncrementalCandidates`, which compares them against ISO timestamps via string sort. All `2026-...` timestamps are lexicographically *less than* `"30m"` (`'2' < '3'`) and `"24h"` (`'20' < '24'`), so `isChanged()` always returned `false` and the pool was silently emptied rather than filtered.

**Fix**: `parseSinceToIso()` normalizes human duration strings to absolute ISO timestamps before the comparison. The log message also shows the resolved timestamp for transparency.

**Verified with local binary**:
- `incrementalSince: "30m"` → 6 memories / 1 chunk (touched file + graph neighbors), all `captureMode: hot`, LLM skipped
- `incrementalSince: "24h"` → 43 memories / 2 chunks, LLM ran, 14 operations found — correct proportionality

PR #654, opened 2026-06-23
fix(improve): build reflect/distill cooldown maps over lane-union scope (#653)

### LINKED ISSUE #653: improve/salience: reflect cooldown guard is ineffective for high-salience/proactive/high-retrieval lanes (candidateRefs scope gap)
## Summary

The once-per-asset reflect cooldown guard (`!lastReflectProposalTs.has(r.ref)`, #643) is **silently ineffective for the high-salience, proactive, and high-retrieval lanes** because the `lastReflectProposalTs` map is built only over `candidateRefs` (= `postCleanupRefs`), while those lanes select refs from a broader pool that can lie outside `candidateRefs`. For any such ref the map has no key, so `!has(ref)` is vacuously `true` and the guard never blocks — no matter how recently the ref was reflected.

## Evidence (production)

`agent:skills/ai-skills/llm-delegation-tool/examples/agents/lore-writer` on prod `~/akm` (beta.41, 2026-06-22):
- It has reflect proposals (most recent `2026-06-20T19:21`) **and 57 `reflect_invoked` events on 2026-06-22**, so `lastReflectProposalTs` *should* contain it and the guard *should* block it.
- It was still selected by the **high-salience lane 57×** in one day; **52/57** were wasted on a downstream reflect `reason=cooldown` (LLM call made, no proposal produced).
- It only stopped when the **no-op dampener** (`consecutive_no_ops=5` → effectiveScore `0.21→0.106`) demoted it below the 10% high-salience cap — an accidental, slow mitigation, not the cooldown guard.

## Root cause (file:line)

`src/commands/improve/improve.ts`:
- `2769`: `const candidateRefs = postCleanupRefs.filter(...).map(r => r.ref);`
- `2771`: `const lastReflectProposalTs = buildLatestProposalTsMap(candidateRefs, "reflect");`
- `buildLatestProposalTsMap` (`882`) filters events by `refSet.has(ref)` (`885`,`890`) — so the map only has keys for refs in `candidateRefs`.
- High-salience gate `3127`: `if (row && row.encoding_salience >= salienceThreshold && !lastReflectProposalTs.has(r.ref))` — iterates `noFeedbackCandidates`, which can include refs **not** in `candidateRefs`. For those, `!has(ref)` is always true → guard never fires.
- Same class of gap at the post-lock proactive re-filter (`1574`, `buildLatestProposalTsMap(proactiveRefStrs, "reflect")`) and the P0-A high-retrieval gate (`3011`, `!lastReflectProposalTs.has(r.ref)`) — both rely on a map scoped to a ref set narrower than the lane they guard.

So the guard works for the signal-delta lane (whose refs ARE `candidateRefs`) but is a no-op for the no-feedback lanes that actually depend on it.

## Impact

- Wasted local-LLM reflect calls: a never-changing high-salience asset is re-selected every run (here 57/day, 52 of them immediately cooled). At scale this is silent compute waste and noise in the high-salience lane.
- The #643 fix is only partially effective; the lanes it was meant to protect are exactly the ones with the gap.

## Proposed fix

Build the reflect/distill cooldown maps over a ref set that covers **every lane the planner will consider**, not just `candidateRefs`. Concretely:

1. Compute the union of all candidate refs the planner may select — `candidateRefs ∪ noFeedbackPool/noFeedbackCandidates ∪ high-retrieval ∪ proactive ∪ high-salience candidates` — BEFORE building `lastReflectProposalTs` / `lastDistillProposalTs`, and build the maps over that union. Then every gate's `!has(ref)` is meaningful.
2. OR, at the high-salience / proactive / high-retrieval gates, look up the latest `reflect_invoked` ts for the specific ref directly (not via the candidateRefs-scoped map) — e.g. a small helper `hasRecentReflect(ref)` querying events per-ref, or a map built over the gate's own iteration set.

Option 1 is preferred (single map, no per-ref query in the hot loop). Ensure the post-lock re-filter (`1574`) and P0-A (`3011`) consume the same union-scoped map.

## Acceptance criteria

- A ref with a recent `reflect_invoked` event is NOT re-selected by the high-salience / proactive / high-retrieval lanes while within the cooldown window, even if it is not in `candidateRefs`.
- Regression: signal-delta lane behavior unchanged.
- A test that reproduces the lore-writer case: ref outside `candidateRefs` but with a recent reflect proposal/`reflect_invoked` event is excluded from the high-salience lane.

## References
- `src/commands/improve/improve.ts:882, 1574, 2769-2772, 3011, 3127`
- #643 (the cooldown guard this completes), #608 (high-salience admission), working-ref §5 F2.

itlackey:
Closes #653

## The bug

The once-per-asset reflect cooldown guard `!lastReflectProposalTs.has(ref)` (#643) is **silently ineffective for the high-salience, proactive, and high-retrieval (P0-A) lanes**. `lastReflectProposalTs` is built only over `candidateRefs` (`buildLatestProposalTsMap(candidateRefs, "reflect")`), whose `buildLatestProposalTsMap` filters events by `refSet.has(ref)` — so the map only ever has keys for refs in `candidateRefs`. Those no-feedback rescue lanes iterate the broader `noFeedbackCandidates` pool, and for any gated ref absent from the candidateRefs-scoped map, `!has(ref)` is vacuously `true`, so the guard never blocks — no matter how recently the ref was reflected.

Production proof (beta.41): `lore-writer` had **57 `reflect_invoked` events in one day** and a reflect proposal from 2026-06-20, yet the high-salience lane re-selected it **57×** (52 wasted on a downstream reflect `reason=cooldown`). It only stopped via the no-op dampener, not the guard.

## The fix (Option 1 — union scope)

Rebuild the reflect/distill cooldown maps over the **union of every lane the planner may select** — `candidateRefs ∪ noFeedbackCandidates` — right after that pool is known, before any rescue-lane gate consumes them:

- `lastReflectProposalTs` / `lastDistillProposalTs` are now `let`, built first over `candidateRefs` (for the signal-delta partition loop, which runs first), then rebuilt over the union once `noFeedbackCandidates` exists.
- `buildLatestProposalTsMap` is a per-ref max over all `*_invoked` events, so the union-scoped map yields **identical values for every `candidateRefs` entry** — the union only ADDS keys for the rescue-lane refs. (Rebuild is skipped entirely when the union equals `candidateRefs`.)
- No change to `buildLatestProposalTsMap`'s logic; passing the union is sufficient.

### Gate sites that now share the union-scoped map

- **High-salience admission** (`improve.ts:3127`, `!lastReflectProposalTs.has(r.ref)`)
- **P0-A high-retrieval** (`improve.ts:3011`, `!lastReflectProposalTs.has(r.ref)`)
- **Proactive maintenance** (`selectProactiveMaintenanceRefs` `lastReflectTs`/`lastDistillTs` args, `improve.ts:3049-3050`)
- **Signal-delta partition loop** (`improve.ts:2813-2814`) — runs before the rebuild with the candidateRefs-scoped map; verified unchanged.

### Post-lock proactive re-filter (`improve.ts:1574`)

Left as-is (with a clarifying comment): it builds a **fresh** map scoped to exactly the refs it filters (`proactiveRefStrs`), so its key set equals its filter set and `!has(ref)` is never vacuously true. The scope-gap that affected the planning gates does not apply there.

## Signal-delta lane is unchanged

The signal-delta lane's refs are exactly `candidateRefs ⊆ union`. Since the map is a per-ref max over all `*_invoked` events, a superset refSet produces byte-identical `has()`/value results for those refs. Confirmed by the existing signal-delta eligibility tests (all green) and a dedicated assertion in the new test (`memory:rated` still reflected).

## Test

Added `#653: high-salience ref with recent reflect is blocked even alongside other lanes; sibling without reflect still fires` to `tests/commands/improve/improve-eligibility.test.ts`. It runs a mixed population — a feedback-bearing signal-delta ref plus two high-salience refs (one with a prior `reflect_invoked`, one without) — and asserts the cooled high-salience ref is blocked, the never-reflected sibling is still admitted (guard doesn't over-block), and the signal-delta ref is unchanged.

**Note on the gap's current reachability:** in the present single-call code path `noFeedbackCandidates ⊆ candidateRefs`, so the gap is **latent** rather than reproducible at unit scope today (verified: the new test passes both with and without the source change). The production divergence is real, and this fix makes the guarantee **structural** — any future lane drawing from a wider pool can no longer silently reintroduce the gap, and every gate's `!has(ref)` is now meaningful by construction. The test pins the guard's correct end-state across the union.

## `bun run check` result

- `lint` — clean (biome + isolation + license + runtime-boundary)
- `tsc --noEmit` — 0 errors
- `test:unit` — 5190 pass / 0 fail
- `test:integration` — 1535 pass / 0 fail

No flakes observed.

## Docs

- `CHANGELOG.md` `[Unreleased] → Fixed`
- `docs/design/improve-salience-working-reference.md` §5 F2 (fix note + issue table row)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **`improve` reflect cooldown guard now covers the no-feedback rescue lanes
+  (#653).** The once-per-asset reflect cooldown (`!lastReflectProposalTs.has(ref)`,
+  #643) was built only over `candidateRefs`, while the high-salience, proactive,
+  and high-retrieval (P0-A) lanes select from a broader pool. A gated ref absent
+  from the candidateRefs-scoped map made `!has(ref)` vacuously true, so the guard
+  silently never fired for those lanes — the lore-writer case re-selected the same
+  high-salience asset 57× in one day (52 wasted on the downstream reflect
+  cooldown). The reflect/distill cooldown maps are now rebuilt over the **union**
+  of every lane's candidate refs (`candidateRefs ∪ noFeedbackCandidates`) before
+  any gate consumes them, so the P0-A, proactive, and high-salience gates all
+  share a map whose `!has(ref)` lookup is meaningful. The signal-delta lane is
+  unchanged (its refs are a subset of the union; the map is a per-ref max over all
+  `*_invoked` events, so existing keys keep identical values).
+
 - **Auto-sync no longer refuses to commit akm's own changes when unrelated
   non-akm files are present in the stash working tree.** When a stash root is
   shared with a project repo, stray files written into the stash root (e.g. a
--- a/docs/design/improve-salience-working-reference.md
+++ b/docs/design/improve-salience-working-reference.md
@@ -288,11 +288,24 @@ never-content-scored refs. (improve.ts ~3375; salience.ts `upsertAssetSalience` 
 `isContentEncodingRow`; distill.ts:973; state-db.ts migration 015)
 
 **F2 — the high-salience gate fires exactly once, then never again.**
-`!lastReflectProposalTs.has(r.ref)` (improve.ts:3126, the #643 cooldown) permanently
+`!lastReflectProposalTs.has(r.ref)` (improve.ts, the #643 cooldown) permanently
 excludes a ref after its first reflect proposal. Auto-accept emits `promoted`, not
 `feedback`, so the ref never re-enters via signal-delta. It must accumulate
 `retrievalCount>=5` to return via high-retrieval. Newly-distilled high-salience assets
 get one pass, then freeze out.
+
+> **#653 (FIXED):** the #643 cooldown map was scoped only to `candidateRefs`, but
+> the high-salience / proactive / P0-A lanes iterate the broader
+> `noFeedbackCandidates` pool. For a gated ref outside `candidateRefs` the map had
+> no key, so `!has(ref)` was vacuously true and the F2 guard never fired for those
+> lanes (lore-writer re-selected 57× in one day). `lastReflectProposalTs` /
+> `lastDistillProposalTs` are now rebuilt over the **union** of every lane's
+> candidate refs (`candidateRefs ∪ noFeedbackCandidates`) right after that pool is
+> known, before the P0-A / proactive / high-salience gates consume them. The map is
+> a per-ref max over all `*_invoked` events, so the signal-delta lane (whose refs
+> are a subset) is byte-identical; the union only ADDS keys for the rescue-lane
+> refs. The post-lock proactive re-filter (improve.ts:1574) already builds its map
+> over exactly the refs it filters, so it had no gap and was left as-is.
 
 **F3 — the no-op dampener can't fire on a churning ref.**
 `resetConsecutiveNoOps` is called on `reflectResult.ok` (improve.ts:4335) = "a proposal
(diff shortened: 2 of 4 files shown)


itlackey:
Closing: a debate-panel review + direct verification found this is a **functional no-op**. `noFeedbackCandidates ⊆ candidateRefs` is provably always true — `eligibleRefs`, `distillOnlyRefs`, and `noFeedbackPool` are all built inside the `for (const r of postCleanupRefs)` loop that does `if (validationFailureRefs.has(r.ref)) continue;` (improve.ts:2826-2827), and `processableRefs = [...eligibleRefs, ...distillOnlyRefs]` (:2909). So `cooldownUnionRefs.length > candidateRefs.length` can never be true and the maps are never rebuilt. The real lore-writer cause was the **stale pre-#643 dist** running most of the day (no cooldown guard at all), not a candidateRefs scope gap — selections collapsed 47→6→1→3→0 once the correct dist deployed. The real lever to stop type-stub-agent selection is the high-salience admission gate (require `isContentEncodingRow`), tracked separately.

Issue #665, opened 2026-06-25
Test infra: purify the integration tier's real-DB fd churn so it can run --parallel>1 (follow-up to #664)


itlackey:
Follow-up to #664. #664 made the **unit** tier fully fd-pure (every real-file `openDatabase()` redirects to a pooled `:memory:` DB under `AKM_TEST_DB_INMEMORY=1`), which eliminated the Bun `--isolate` epoll fd race there and let `test:unit` flip to `--parallel=4` (20/20 clean gate). The **integration** tier was intentionally left out of scope and still runs at `--parallel=1` with the epoll retry wrapper.

## Why integration still churns fds
The integration tier (`tests/integration/`, plus the double-run `tests/commands` / `tests/workflows`) does **genuine real-DB / real-process I/O** — that's its purpose. It includes:
- The ~10 real-DB tests #664 relocated there because they test on-disk behavior: `sqlite-journal-mode` (WAL/journal pragmas), `proposal-storage-sqlite` (WAL concurrency via Worker subprocesses), `logs-db` (cross-DB `ATTACH`), `index-db-loan.characterization` (cross-engine loan lifecycle), `improve-db-locking` (`SQLITE_BUSY`/locking), `workflows/migrations` (migration-ledger bootstrap from a raw-seeded file), `session-logs-providers` (external `opencode.db`/claude readers), `purity-guard` (the guard itself), `issues-191-194` (WAL-preserved), `fixtures-stash-load` (asserts `index.db` lands on disk).
- All the subprocess-bound tests relocated in #664 (real `git`/`tar`/`docker`/`bash`/lock/PID).

Because of this real fd + `-wal`/`-shm` churn, the epoll race can still surface there, so `run-test-shard.sh` keeps the full retry wrapper (race-signature grep + hang detection + 3-attempt loop) **only** on the integration path, and the default stays `--parallel=1`.

## Scope of this issue
Make the integration tier safe to run `--parallel>1` (and ideally delete the retry wrapper entirely), by one or more of:

1. **Per-subsystem `:memory:`/in-process seams** for the genuinely-DB tests that don't actually need on-disk semantics — several of the relocated files only need a real file because of an incidental check, not their true subject. Audit each for a `db?`/`:memory:` path (the `seedEntries` / `searchOnDb` / `openStateDatabase(":memory:")` seams from #664 already exist).
[...]

PR #666, opened 2026-06-25
Eliminate the Bun --isolate epoll fd race + flip the unit suite to --parallel=4 (#664)

### LINKED ISSUE #664: Test infra: eliminate the Bun --isolate epoll fd race + cut unit-suite runtime (follow-up to #663)
Follow-up to #663. Two related problems sharing one root cause (real-subprocess / server fd churn in the unit suite). #663 made CI **green** by self-healing the race in the retry wrapper, but the underlying Bun runtime bug and the suite slowness remain.

### What happens
Bun 1.3.x `bun test` (even at `--parallel=1`) intermittently hits an fd/epoll race: an fd is recycled into a still-registered epoll slot. **Two manifestations:**
- **Hang** — a worker parks in `ep_poll`; `timeout` kills it (exit 124/137).
- **Crash** — runtime aborts with `error: EEXIST: file already exists, epoll_ctl`, the runner then logs `Cannot call beforeEach() after the test run has completed`, and the process exits with an ordinary code (1).

### Observed
Release run `28067662432` failed **deterministically twice** on `test-unit (5/8)` — `tests/core/write-source.test.ts` (heavy `spawnSync` git: init/commit/push). That file passes clean locally and on every other shard; **no assertion failed** — the runtime crashed. The crash exit code (1) fell through the wrapper's hang-only (124/137) retry branch and was misreported as a real test failure.

### Current mitigation (#663, shipped beta.45)
`scripts/run-test-shard.sh` now `tee`s each attempt's output and retries on a fresh process for **both** the hang exit codes **and** the race signature (`EEXIST.*epoll_ctl` / `Cannot call (before|after)(Each|All)() after the test run has completed`). These strings are emitted only by the Bun runtime, never by a failed `expect()`, so a genuine red test (no signature + non-timeout exit) still fails fast on attempt 1 and is never masked; if the race never clears across 3 attempts the build stays red.

### Why this is still a bug to fix
Self-healing ≠ eliminated. The trigger is `spawnSync`/`Bun.serve` fd churn, and **47 unit-test files** spawn subprocesses, so the race can surface in any shard. True fixes to evaluate:
- Reduce fd churn in the unit tree — move real-`spawnSync`/real-`Bun.serve` tests to `tests/integration/` per the existing unit/integration boundary policy (start with the heaviest: `write-source.test.ts`, the stdin-driven `spawnSync` tests).
- Investigate whether a Bun upgrade or a `bun test` flag (isolate behavior) removes the epoll collision.
- If eliminated, `--parallel>1` becomes safe again → large CI + local speedup (see §2 long tail).

### Measurement caveat
`bun run test:time` does **not** exclude `tests/integration/`, but `test:unit` does — ~111s of the timing report is integration files that never run in `test:unit`. **Fix the timing script** (`scripts/test-timing-report.ts`) to apply `--path-ignore-patterns=tests/integration` so numbers reflect reality.

### Top non-integration offenders (per-file, includes ~270ms Bun startup artifact)
| ms | file | root cause |
|---|---|---|
| 19,810 | `tests/workflow-cli.test.ts` | 108 `runCli()` + 64 `createWorkflowEnv()` × 5 `mkdtempSync` (~320 temp dirs) + SQLite churn |
| 9,920 | `tests/source.test.ts` | 5 inline full `akmIndex({full:true})`, fresh stash per test |
| 8,200 | `tests/registry-search.test.ts` | a fresh `Bun.serve` + SQLite cache DB **per test** (40×) |
| 7,600 | `tests/curate-command.test.ts` | 15 invocations, each auto-indexes; 3 copy a 22-file fixture |
| 5,070 | `tests/commands/improve-memory-misc.test.ts` | **16 full `akmIndex` rebuilds** for 20 tests |
| 4,830 | `tests/commands/improve/consolidate-min-pool-size.test.ts` | 4 full `akmIndex` rebuilds |
| 4,820 | `tests/init.test.ts` | 3 `mkdtempSync`/test + PATH scan + 10 file copies × 14 tests |
| 3,170 | `tests/issues-191-194.test.ts` | multiple sequential in-process CLI chains; search auto-indexes |
| 3,070 | `tests/env-path-run.test.ts` | 8 real `spawnSync("bun", …)` (stdin tests) |
| 2,600 | `tests/health-command.test.ts` | 6 isolated dirs/test × 40 tests (240 `mkdtempSync`) |

Real wall-clock sleeps to remove: `llm-client.test.ts` (460ms), `llm-feature-gate.test.ts` (2×200ms), `events.test.ts` (~110ms).

### Prioritized fixes (low risk → behavior-changing)
**P1 (low risk, high impact):**
- Exclude `tests/integration` from `test-timing-report.ts` (accurate numbers).
- `beforeAll`-share one index build in the `akmIndex`-heavy files (`improve-memory-misc`, `source`, `consolidate-min-pool-size`) instead of per-test rebuilds.
- Share one `Bun.serve` (constant `FIXTURE_INDEX`) in `registry-search.test.ts` via `beforeAll`/`afterAll`; keep per-test XDG isolation.
- Replace the full-CLI `Bun.spawnSync` in `tests/fixtures/stashes/load.ts` with in-process `akmIndex` (the `skipIndex` path already exists).

**P2 (moderate):**
- Group read-only tests per `describe` to share envs in `workflow-cli.test.ts` (reduce ~320 `mkdtempSync`).
- Replace real `setTimeout` waits with injected-clock/fake-timer in `llm-client.test.ts` / `llm-feature-gate.test.ts`.
- Audit `env-path-run.test.ts` — move non-stdin `spawnSync` calls to in-process `runCliCapture`.

**P3:** the 336-file long tail (~150s) is strictly sequential because of `--parallel=1`. The real win there is gated on §1 (make parallelism safe). Justified spawns (`completions.test.ts`, `events.test.ts` exec-boundary) should stay.

### Acceptance
- `test-timing-report.ts` excludes integration.
- The heaviest `akmIndex`/`Bun.serve` per-test rebuilds are shared; measurable `test:unit` reduction.
- A decision recorded on §1 (relocate fd-churning tests and/or Bun upgrade) so `--parallel>1` can be re-evaluated.

Refs: #663 (wrapper fix), `akm-bun-parallel-test-hang` memory, unit/integration boundary policy.

itlackey:
Closes #664. Follow-up tracked in #665.

## What this does
Eliminates the Bun `--isolate` `epoll_ctl EEXIST` fd race in the unit test tier and flips `test:unit` to `--parallel=4`, cutting unit-suite wall time ~3× (≈110–150s → ~47s). Both halves of #664 are delivered.

## The core insight
The epoll race and the slow suite were **one symptom**: the production code had no I/O seams, so unit tests stood up real subprocesses / `Bun.serve` servers / on-disk SQLite DBs — and that fd churn is what triggers the race under parallelism. The fix is **dependency seams + one structural DB redirect**, not test-harness workarounds, and explicitly **not** 900 hand-migrations.

## How it's structured (seam by seam)
- **Seam 1 — HTTP (`HttpClient`):** injectable `fetch` threaded through `fetchWithTimeout`/`fetchWithRetry`, the LLM client, embedder, registry providers, and graph-extract. ~26 `Bun.serve` instances across ~17 files collapsed to injected fakes (zero sockets). `ALLOWED_SERVE` → 0.
- **Seam 2 — `:memory:` entries + guard:** `GetAllEntries` entry-source + `seedEntries` (real `upsertEntry`/`getAllEntries` on a `:memory:` DB, zero query drift); improve planner reads via the injectable source; opt-in runtime purity guard at the storage boundary; do-now fixes (issue C isolation rethrow, issue A sink reset, issue G deterministic readdir).
- **Seam 3 — search + issue E:** `searchOnDb(db)` / `db?` injection; removed the in-flight `process.env.AKM_DISABLE_PROJECT_CONTEXT` mutation + ambient `cwd` read from the search hot path (gate precondition).
- **Seam 4 — stdin:** guarded `StdinPort`; stdin spawns migrated to `runCliCapture`; `env run`/`secret run` relocated to integration.
- **Seam 5 — time + F-min:** injectable `now`/interval + `ctx.db` into `readEvents` (also fixed a latent improve bug); `env` param threaded through the path resolvers (gate precondition); real sleeps virtualized.
- **The structural fix that made the flip possible:** under the unit tier (`AKM_TEST_DB_INMEMORY=1`), `openDatabase(path)` redirects to a process-pooled `:memory:` DB keyed by path (close = keep-alive, LRU-capped), with a single pool-aware `databaseExists()` replacing the ~9 `fs.existsSync(dbPath)` guards. **One redirect** made the whole unit tier fd-pure. The ~10 genuinely-real-DB tests (WAL/journal/locking/migrations/external readers) moved to the integration tier.

## Guardrails
- `scripts/lint-tests-unit-purity.ts`: shrink-only ratchet (real-serve / real-spawn / `akmIndex({full:true})`-in-test-body) — now at **baseline 0**, with a meta-test that pins it.
- `scripts/run-test-shard.sh`: `SHARD_PARALLEL` knob; unit default → 4, the epoll retry wrapper **deleted from the unit path** (kept on integration, which still has real-DB churn — see #665).

## The flip gate (honest, empirical)
- Static baseline 0 is necessary but not sufficient — a global runtime-guard run revealed ~900 real DB opens the lint can't see. The structural pool fixed that at the source.
- 20× `--parallel=4` gate: **20/20 clean, 0 fail, 0 race-signature, 0 hang** (it was 2/20 race-or-hang before the pool). The 3 tests that flaked under load were fixed at root cause (path collision, wall-clock race, stale cache key) — no thresholds loosened.

## Verification
- `bun run lint` 0/0 · `bunx tsc --noEmit` 0
- `test:unit` **4717 pass / 0 fail** at `--parallel=4` (~47s)
- `test:integration` **1947 pass / 0 fail** at `--parallel=1`
- Design docs: `docs/technical/test-io-seam-map.md`, `test-io-seam-design.md` (+ §8 addendum), `test-io-seam-execution-plan.md`

## Out of scope (→ #665)
The integration tier still runs `--parallel=1` with the retry wrapper because it does genuine real-DB / real-process I/O. Purifying / sharding that tier is tracked in #665.


itlackey:
Addressed the Copilot review in `904dbc3c`:

- **`remote.ts:67` / `remote.ts:111` (uncancellable embedding)** — fixed. `RemoteEmbedder.embed`/`embedBatch` now pass the caller's `AbortSignal` as `fetchWithTimeout`'s **4th positional arg** (the seam that links it to the timeout controller) instead of `RequestInit.signal`, which `fetchWithTimeout` overwrites. Dropped the dead `opts.signal`. Remote embedding is now genuinely cancellable. (Good catch — this was a latent bug the seam change surfaced.)
- **`seed-entries.ts:58` (layout divergence)** — fixed. `seedEntries` now derives the directory from the canonical `TYPE_DIRS` map (`memory`→`memories`, …) instead of naive `<type>s`, so seeded `filePath`s match AKM's real on-disk layout and the planner's path/existence guards behave as against a real stash. Falls back to `<type>s` only for an unregistered type.
- **`seed-entries.test.ts:31`** — updated the assertion to the canonical `memories/` path.

Verified: `tsc` 0, `bun run lint` 0/0, `test:unit` 4717/0 at `--parallel=4`.

itlackey:
Superseded by #667 (merged): #664 was fixed via process-level sharding (scripts/test-unit.sh runs N shards as separate processes, each --parallel=1), which is race-free by construction and ~3× faster than single-process. The in-process --parallel=4 approach here is no longer needed. Closing and removing the branch.

PR #715, opened 2026-07-12
Implement stash organization conventions (SPEC-1..8)

itlackey:
## Summary

This PR implements the stash organization and back-linking conventions (SPEC-1 through SPEC-8) as defined in `docs/design/stash-organization-conventions.md` and `docs/design/stash-conventions-code-spec.md`. The changes enable agents to organize assets hierarchically, cross-link them for improved retrieval, and maintain referential integrity across the stash.

### Key Features

**SPEC-1: Frontmatter xref channels in lint**
- Extended `akm lint` missing-ref check to scan `xrefs:`, `supersededBy:`, and `contradictedBy:` frontmatter keys
- Dangling refs in these channels are now flagged with detail naming the key
- Applies to all markdown asset types (memories, knowledge, lessons, facts, agents, commands, skills, workflows)

**SPEC-3: Cross-references on write (`--xref`)**
- Added `--xref <ref>` repeatable flag to `akm remember` and `akm import` commands
- Validates refs before any write happens; unresolvable refs cause exit 2 with standard error envelope
- Merges xrefs into existing frontmatter on import (deduplicating)
- Soft cap of 5 xrefs with stderr warning (non-blocking)
- Written xrefs fold into FTS hints for discoverability

**SPEC-4: Ref-prefix enumeration in search**
- Implemented `<type>:<prefix>/` and bare `<type>:` query syntax for typed subtree enumeration
- `akm search "memory:projectA/"` now lists exactly that subtree (recursive, `/`-boundary exact)
- Bare `memory:` enumerates the whole type
- Composes with belief/source filters and limit

**SPEC-5: Supersession with `--supersedes`**
- Added `--supersedes <ref>` flag to `akm remember` and `akm import`
- Atomically writes new correction asset with old ref in `xrefs:` AND demotes old asset to `beliefState: superseded` with `supersededBy:` list
- Demotion is immediately live via reindexing; beliefStateBoost ranks corrections above stale incumbents
- Idempotent: re-running doesn't duplicate supersededBy entries
- Git targets batch both files into single boundary commit

**SPEC-6: Fact category capture**
- Added `category?: string` field to StashEntry for frontmatter `category:` key
- Captured in `applyCuratedFrontmatter` and whitelisted through `validateStashEntry`
- Survives .stash.json round-trip for rank-time and filter policies
- Convention facts marked with `category: convention` in shipped stash skeleton

**SPEC-7: Asset rename with xref rewriting (`akm mv`)**
- Implemented `akm mv <ref> <new-name>` command for renaming assets within type directories
- Rewrites inbound refs across writable stash's markdown files (body, frontmatter keys, fenced code blocks)
- Uses complete-ref boundary matching to avoid rewriting longer refs sharing old ref as prefix
- Memory `.derived.md` twins move together, preserving `entry_key === <base>.derived` coupling
- Re-keys index row IN PLACE to preserve utility_scores history
- Read-only sources scanned but never written; reported in `readOnlyCiters` for manual follow-up
- Appends `mv` event to state.db events stream on success

**SPEC-8: Ref-prefix search with body-opening index**
- Added optional `indexBodyOpening` config to index first paragraph of assets
- Enables better ranking for ref-prefix queries by including opening context in search fields

### Supporting Changes

- Added convention facts to stash skeleton: `organization.md`, `backlinks.md`, `domains.md`
- Updated asset-type convention facts with maintenance strategies
- Extended frontmatter mutation utilities for belief-state transitions
- Added comprehensive test suites for all specs (759 lines for mv, 614 for supersedes, 522 for xref, 363 for ref-prefix search, 245 for

https://claude.ai/code/session_01QvPuW7VMxWgmeQTm8RxK78

### DIFF
--- a/src/commands/improve/memory/memory-belief.ts
+++ b/src/commands/improve/memory/memory-belief.ts
@@ -39,6 +39,26 @@ export type {
   MemoryBeliefTransitionLogRecord,
 } from "./memory-improve";
 
+// ── Shared edge-list reader ───────────────────────────────────────────────────
+
+/**
+ * Read a frontmatter edge value (`supersededBy` / `contradictedBy`) as a
+ * string list, promoting a scalar string to a one-element list.
+ *
+ * Scalar edges are LIVE data: the indexer's `normalizeNonEmptyStringList`
+ * accepts them and lint deliberately never flags them. An Array.isArray-only
+ * read treats a scalar as "no existing edges" and silently destroys the edge
+ * on the next merge — mirror `mergeXrefsIntoContent`'s scalar promotion
+ * instead.
+ */
+function readEdgeList(value: unknown): string[] {
+  if (Array.isArray(value)) {
+    return value.filter((v): v is string => typeof v === "string" && v.trim().length > 0);
+  }
+  if (typeof value === "string" && value.trim()) return [value.trim()];
+  return [];
+}
+
 // ── Contradiction edge writer ─────────────────────────────────────────────────
 
 /**
@@ -58,9 +78,7 @@ export type {
  */
 export function writeContradictEdge(filePath: string, contradictedByRef: string): void {
   mutateFrontmatter(filePath, (parsed) => {
-    const existing: string[] = Array.isArray(parsed.data.contradictedBy)
-      ? (parsed.data.contradictedBy as string[])
-      : [];
+    const existing = readEdgeList(parsed.data.contradictedBy);
     if (existing.includes(contradictedByRef)) return null; // Already written — idempotent.
 
     const nextContradictedBy = [...new Set([...existing, contradictedByRef])].sort();
@@ -89,13 +107,22 @@ export function writeContradictEdge(filePath: string, contradictedByRef: string)
  * is already demoted, the file is not rewritten. Multiple corrections
  * sorted-set-append their refs.
  *
+ * Never WEAKENS an existing demotion: `contradicted` and `archived` rank
+ * BELOW `superseded` (severity order deprecated > superseded > contradicted >
+ * archived — see `BELIEF_STATE_SCORE_CEILINGS` in
+ * src/indexer/search/ranking-contributors.ts), so superseding an already
+ * contradicted/archived asset keeps the stronger state and only appends the
+ * `supersededBy` edge.
+ *
  * @param filePath        - Absolute path to the asset markdown file.
  * @param supersededByRef - The ref of the correction that supersedes this asset.
  */
 export function writeSupersededEdge(filePath: string, supersededByRef: string): void {
   mutateFrontmatter(filePath, (parsed) => {
-    const existing: string[] = Array.isArray(parsed.data.supersededBy) ? (parsed.data.supersededBy as string[]) : [];
-    if (existing.includes(supersededByRef) && parsed.data.beliefState === "superseded") {
+    const existing = readEdgeList(parsed.data.supersededBy);
+    const currentState = parsed.data.beliefState;
+    const nextState = currentState === "contradicted" || currentState === "archived" ? currentState : "superseded";
+    if (existing.includes(supersededByRef) && currentState === nextState) {
       return null; // Already written — idempotent.
     }
 
@@ -103,7 +130,7 @@ export function writeSupersededEdge(filePath: string, supersededByRef: string): 
     return {
       ...parsed.data,
       supersededBy: nextSupersededBy,
-      beliefState: "superseded",
+      beliefState: nextState,
     };
   });
 }
--- a/src/commands/improve/memory/memory-belief.ts
+++ b/src/commands/improve/memory/memory-belief.ts
@@ -71,3 +71,39 @@ export function writeContradictEdge(filePath: string, contradictedByRef: string)
     };
   });
 }
+
+// ── Supersession edge writer ─────────────────────────────────────────────────
+
+/**
+ * Write `supersededBy` and `beliefState: superseded` edges to an asset file's
+ * frontmatter (SPEC-5, stash-conventions-code-spec.md).
+ *
+ * Sibling of {@link writeContradictEdge}: the shared primitive for the
+ * conventions' corrections pattern — when a new asset supersedes an old one,
+ * the old asset gets a METADATA-ONLY demotion edit (every other frontmatter
+ * key and the body are preserved) so the ranker's beliefStateBoost demotes the
+ * stale incumbent and `--belief current` hides it. Used by
+ * `akm remember --supersedes` / `akm import --supersedes`.
+ *
+ * Idempotent: if `supersededByRef` is already in `supersededBy` and the file
+ * is already demoted, the file is not rewritten. Multiple corrections
+ * sorted-set-append their refs.
+ *
+ * @param filePath        - Absolute path to the asset markdown file.
+ * @param supersededByRef - The ref of the correction that supersedes this asset.
+ */
+export function writeSupersededEdge(filePath: string, supersededByRef: string): void {
+  mutateFrontmatter(filePath, (parsed) => {
+    const existing: string[] = Array.isArray(parsed.data.supersededBy) ? (parsed.data.supersededBy as string[]) : [];
+    if (existing.includes(supersededByRef) && parsed.data.beliefState === "superseded") {
+      return null; // Already written — idempotent.
+    }
+
+    const nextSupersededBy = [...new Set([...existing, supersededByRef])].sort();
+    return {
+      ...parsed.data,
+      supersededBy: nextSupersededBy,
+      beliefState: "superseded",
+    };
+  });
+}
--- a/src/commands/improve/memory/memory-belief.ts
+++ b/src/commands/improve/memory/memory-belief.ts
@@ -70,8 +70,18 @@ function readEdgeList(value: unknown): string[] {
  *   - `memory-contradiction-detect.ts` for the M-1 automated contradiction pass
  *   - `resolveFamilyContradictions` in `memory-improve.ts` for SCC resolution
  *
- * Idempotent: if the `contradictedByRef` is already in `contradictedBy`,
- * the file is not rewritten.
+ * Idempotent: if the `contradictedByRef` is already in `contradictedBy` AND
+ * the file already carries the demotion state, the file is not rewritten. The
+ * guard is state-aware like its sibling {@link writeSupersededEdge}: an edge
+ * present WITHOUT the demotion (e.g. a hand-written `contradictedBy:` line,
+ * or a beliefState lost to a partial edit) is repaired, not skipped — an
+ * edge-only guard would make such a file a permanent no-op while
+ * consolidate's handleContradictOp reports the contradiction as applied.
+ *
+ * Never weakens a stronger demotion: `archived` ranks BELOW `contradicted`
+ * (see `BELIEF_STATE_SCORE_CEILINGS` in
+ * src/indexer/search/ranking-contributors.ts), so contradicting an archived
+ * memory keeps `archived` and only appends the edge.
  *
  * @param filePath          - Absolute path to the memory markdown file.
  * @param contradictedByRef - The ref that contradicts this memory.
@@ -79,13 +89,17 @@ function readEdgeList(value: unknown): string[] {
 export function writeContradictEdge(filePath: string, contradictedByRef: string): void {
   mutateFrontmatter(filePath, (parsed) => {
     const existing = readEdgeList(parsed.data.contradictedBy);
-    if (existing.includes(contradictedByRef)) return null; // Already written — idempotent.
+    const currentState = parsed.data.beliefState;
+    const nextState = currentState === "archived" ? currentState : "contradicted";
+    if (existing.includes(contradictedByRef) && currentState === nextState) {
+      return null; // Already written — idempotent.
+    }
 
     const nextContradictedBy = [...new Set([...existing, contradictedByRef])].sort();
     return {
       ...parsed.data,
       contradictedBy: nextContradictedBy,
-      beliefState: "contradicted",
+      beliefState: nextState,
     };
   });
 }
--- a/src/commands/lint/base-linter.ts
+++ b/src/commands/lint/base-linter.ts
@@ -150,12 +150,33 @@ function buildRefTypeAlternation(): string {
   return types.map((t) => t.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")).join("|");
 }
 
+/**
+ * Body-ref boundary grammar, shared with `akm mv`'s ref-rewrite pattern —
+ * `src/commands/mv-cli.ts` imports these constants so the two grammars cannot
+ * drift. Any character-class change here retargets both the lint missing-ref
+ * scan and mv's inbound-xref rewriting.
+ *
+ * `REF_BOUNDARY_PREFIX_CLASS_SRC` is the character class a ref may start
+ * after (a ref also matches at line start): whitespace, backtick, quote,
+ * `(`, or `[` — the `[` admits markdown-link-style refs like
+ * `see [memory:foo]`, which the legacy class silently skipped.
+ *
+ * `REF_SLUG_CHAR_CLASS_SRC` is the character class a ref's `<type>:<slug>`
+ * token is made of; the first excluded character ends the ref.
+ */
+export const REF_BOUNDARY_PREFIX_CLASS_SRC = "[\\s`\"'(\\[]";
+export const REF_SLUG_CHAR_CLASS_SRC = "[^\\s\"'`)\\]>,\\n]";
+
 // Only the TYPE alternation is registry-derived; the surrounding grammar
 // (boundary prefix, capture group, slug charset) is byte-identical to the
-// legacy hand-written pattern. Deriving the types from `getAssetTypes()` means
-// `env`/`secret` (added in 0.9) are now matched, and the removed `vault` type
-// is not — both follow the registry automatically.
-const REF_RE = new RegExp(`(?:^|[\\s\`"'(])((${buildRefTypeAlternation()}):[^\\s"'\`)\\]>,\\n]+)`, "gm");
+// legacy hand-written pattern, except that the boundary prefix now also
+// admits `[`. Deriving the types from `getAssetTypes()` means `env`/`secret`
+// (added in 0.9) are now matched, and the removed `vault` type is not — both
+// follow the registry automatically.
+const REF_RE = new RegExp(
+  `(?:^|${REF_BOUNDARY_PREFIX_CLASS_SRC})((${buildRefTypeAlternation()}):${REF_SLUG_CHAR_CLASS_SRC}+)`,
+  "gm",
+);
 
 /**
  * Map from ref type to relative path pattern within stashRoot. Returns null to
@@ -390,6 +411,23 @@ function readRefsArray(value: unknown): string[] | null {
     if (typeof entry === "string" && entry.trim()) out.push(entry.trim());
   }
   return out;
+}
+
+/**
+ * Like {@link readRefsArray} but also accepts a single scalar string,
+ * normalizing it to a one-element list. The indexer's
+ * `normalizeNonEmptyStringList` treats `supersededBy: memory:x` and
+ * `supersededBy: [memory:x]` identically — both are live data — so the
+ * frontmatter xref-channel check must validate both shapes; the array-only
+ * reader silently skipped dangling scalar refs. Returns `null` for any other
+ * type (missing key, number, object) and for a blank scalar.
+ */
+function readRefStringOrArray(value: unknown): string[] | null {
+  if (typeof value === "string") {
+    const trimmed = value.trim();
+    return trimmed ? [trimmed] : null;
+  }
+  return readRefsArray(value);
 }
 
 /**
@@ -663,15 +701,22 @@ export abstract class BaseLinter implements AssetLinter {
       // Non-ref-shaped values (URLs, `raw/<slug>`, `<placeholder>`
       // templates, shell vars) fall out via checkMissingRefs' guards.
       //
-      // Skipped when ctx.frontmatter is null: on the task/YAML path
-      // (lint/index.ts) the whole file IS the body (`body === raw`), so
-      // ref values under these keys are already caught by the body scan
-      // above — running the pass again would double-report each dangling
-      // ref. Md files without a frontmatter block also land here, with
-      // empty ctx.data, so nothing is lost for them either.
-      if (ctx.frontmatter !== null) {
+      // Gate: runs when the file has a frontmatter block OR when an
+      // authoritative `refs:` list was extracted. On the task/YAML path
+      // (lint/index.ts) ctx.frontmatter is always null and the whole file
+      // IS the body (`body === raw`); the top-level YAML keys land in
+      // ctx.data. Without `refs:` the body scan above already catches ref
+      // values under these keys, so running the pass would double-report —
+      // skip it. With `refs:` present the body scan is suppressed
+      // (refSource is the refs list), so this pass is the ONLY thing that
+      // validates the xref keys — it must run or dangling task xrefs go
+      // unreported. The two cases are mutually exclusive, so no ref is
+      // ever double-reported. Md files without a frontmatter block and
+      // without `refs:` land in the skip branch with empty ctx.data, so
+      // nothing is lost for them either.
+      if (ctx.frontmatter !== null || explicitRefs !== null) {
         for (const key of XREF_FRONTMATTER_KEYS) {
-          const values = readRefsArray(ctx.data?.[key]);
+          const values = readRefStringOrArray(ctx.data?.[key]);
           if (values === null) continue;
           const missingXrefs = checkMissingRefs(values.join("\n"), ctx.stashRoot, ctx.extraStashRoots);
           for (const { ref, resolvedRelPath } of missingXrefs) {
--- a/src/commands/lint/base-linter.ts
+++ b/src/commands/lint/base-linter.ts
@@ -310,6 +310,21 @@ function checkMissingRefs(
 // ── frontmatter refs ─────────────────────────────────────────────────────────
 
 /**
+ * Frontmatter keys that carry cross-reference lists per the stash
+ * organization conventions: `xrefs:` (provenance / associative links),
+ * `supersededBy:` and `contradictedBy:` (belief-state correction links).
+ * The missing-ref check validates each of these in ADDITION to the body /
+ * `refs:` scan — they are the channel the conventions mandate, and a rename
+ * would otherwise dangle them silently.
+ *
+ * `sources:` is deliberately excluded (non-wiki `sources:` was rejected as a
+ * typed channel; wiki `sources:` is checked by lintWiki). `source_refs:` /
+ * `evidenceSources:` are excluded too — they legitimately point at
+ * merged-away or pruned assets (historical provenance).
+ */
+const XREF_FRONTMATTER_KEYS = ["xrefs", "supersededBy", "contradictedBy"] as const;
+
+/**
  * Return the `refs:` array from frontmatter when it is present and is an
  * array of strings; otherwise return `null` to signal the caller should
  * fall back to scanning the body. An empty array (`refs: []`) is also
@@ -616,6 +631,35 @@ export abstract class BaseLinter implements AssetLinter {
           detail: `missing ref: ${ref} (resolved to ${resolvedRelPath})`,
           fixed: false,
         });
+      }
+
+      // Frontmatter xref channels (xrefs / supersededBy / contradictedBy).
+      // Runs regardless of the `refs:` body-scan carve-out above — that
+      // carve-out governs only the BODY scan (`refs: []` declares "no
+      // outbound refs in the body", not "skip my correction links").
+      // Non-ref-shaped values (URLs, `raw/<slug>`, `<placeholder>`
+      // templates, shell vars) fall out via checkMissingRefs' guards.
+      //
+      // Skipped when ctx.frontmatter is null: on the task/YAML path
+      // (lint/index.ts) the whole file IS the body (`body === raw`), so
+      // ref values under these keys are already caught by the body scan
+      // above — running the pass again would double-report each dangling
+      // ref. Md files without a frontmatter block also land here, with
+      // empty ctx.data, so nothing is lost for them either.
+      if (ctx.frontmatter !== null) {
+        for (const key of XREF_FRONTMATTER_KEYS) {
+          const values = readRefsArray(ctx.data?.[key]);
+          if (values === null) continue;
+          const missingXrefs = checkMissingRefs(values.join("\n"), ctx.stashRoot, ctx.extraStashRoots);
+          for (const { ref, resolvedRelPath } of missingXrefs) {
+            issues.push({
+              file: ctx.relPath,
+              issue: "missing-ref",
+              detail: `missing ref: ${ref} (frontmatter ${key}; resolved to ${resolvedRelPath})`,
+              fixed: false,
+            });
+          }
+        }
       }
     }
 
--- a/src/commands/lint/base-linter.ts
+++ b/src/commands/lint/base-linter.ts
@@ -158,13 +158,19 @@ function buildRefTypeAlternation(): string {
  *
  * `REF_BOUNDARY_PREFIX_CLASS_SRC` is the character class a ref may start
  * after (a ref also matches at line start): whitespace, backtick, quote,
- * `(`, or `[` — the `[` admits markdown-link-style refs like
- * `see [memory:foo]`, which the legacy class silently skipped.
+ * `(`, `[`, or `,` — the `[` admits markdown-link-style refs like
+ * `see [memory:foo]`, which the legacy class silently skipped, and the `,`
+ * admits the ref AFTER a bare comma in a no-space flow list like
+ * `xrefs: [memory:a,memory:b]` (valid YAML). `,` is already a slug
+ * TERMINATOR (excluded from `REF_SLUG_CHAR_CLASS_SRC`), so `a,b` splits
+ * cleanly and adding it here cannot extend any existing match; false
+ * positives are fenced by the type alternation — a comma only starts a match
+ * when a literal `<type>:` follows it.
  *
  * `REF_SLUG_CHAR_CLASS_SRC` is the character class a ref's `<type>:<slug>`
  * token is made of; the first excluded character ends the ref.
  */
-export const REF_BOUNDARY_PREFIX_CLASS_SRC = "[\\s`\"'(\\[]";
+export const REF_BOUNDARY_PREFIX_CLASS_SRC = "[\\s`\"'(,\\[]";
 export const REF_SLUG_CHAR_CLASS_SRC = "[^\\s\"'`)\\]>,\\n]";
 
 // Only the TYPE alternation is registry-derived; the surrounding grammar
--- a/src/commands/lint/base-linter.ts
+++ b/src/commands/lint/base-linter.ts
@@ -194,41 +194,64 @@ export function refToRelPath(refType: string, refName: string): string | null {
  */
 export function refExistsInAnyStash(relPath: string, refType: string, refName: string, stashRoots: string[]): boolean {
   for (const root of stashRoots) {
-    const absPath = path.join(root, relPath);
-    if (fs.existsSync(absPath)) return true;
-    // Multi-file skill layout: directory containing SKILL.md
-    const bareDir = absPath.replace(/\.md$/, "");
-    if (fs.existsSync(bareDir) && fs.existsSync(path.join(bareDir, "SKILL.md"))) return true;
-    // .derived.md variant for memory refs
-    if (refType === "memory") {
-      const derivedPath = path.join(root, "memories", `${refName}.derived.md`);
-      if (fs.existsSync(derivedPath)) return true;
-    }
-    // Knowledge-specific: search subdirectories like knowledge/projects/, knowledge/tools/, etc.
-    if (refType === "knowledge") {
-      try {
-        const knowledgeDir = path.join(root, "knowledge");
-        if (fs.existsSync(knowledgeDir) && fs.statSync(knowledgeDir).isDirectory()) {
-          const entries = fs.readdirSync(knowledgeDir, { withFileTypes: true });
-          for (const entry of entries) {
-            if (!entry.isDirectory()) continue;
-            const subPath = path.join(knowledgeDir, entry.name, `${refName}.md`);
-            if (fs.existsSync(subPath)) return true;
-          }
-        }
-      } catch {
-        // Ignore errors reading directory
-      }
-    }
-    // Fallback: the refName may already encode the full stash-relative path
-    // (e.g. knowledge:skills/foo/references/bar where the file lives at
-    // <stash>/skills/foo/references/bar.md, not <stash>/knowledge/skills/...).
-    const directPath = path.join(root, `${refName}.md`);
-    if (fs.existsSync(directPath)) return true;
-    const directDir = path.join(root, refName);
-    if (fs.existsSync(directDir) && fs.existsSync(path.join(directDir, "SKILL.md"))) return true;
+    if (resolveRefPathInStash(relPath, refType, refName, root) !== null) return true;
   }
   return false;
+}
+
+/**
+ * Resolve the on-disk primary file for a ref within a SINGLE stash root, using
+ * the same reachability rules (in the same order) as
+ * {@link refExistsInAnyStash}, which delegates here. Returns the absolute path
+ * of the file that makes the ref "exist" — for a multi-file skill directory
+ * that is its `SKILL.md` primary — or `null` when the ref does not resolve in
+ * this root.
+ *
+ * Extracted for SPEC-5 (`--supersedes` demotion): write commands need the
+ * superseded asset's actual file to mutate, and forking a second resolver
+ * would drift from lint's. NOT part of the akm-plugins ref-resolver contract
+ * (the contract pins `refToRelPath` + `refExistsInAnyStash`; this is the
+ * shared internal both build on).
+ */
+export function resolveRefPathInStash(relPath: string, refType: string, refName: string, root: string): string | null {
+  const absPath = path.join(root, relPath);
+  if (fs.existsSync(absPath)) return absPath;
+  // Multi-file skill layout: directory containing SKILL.md
+  const bareDir = absPath.replace(/\.md$/, "");
+  if (fs.existsSync(bareDir) && fs.existsSync(path.join(bareDir, "SKILL.md"))) {
+    return path.join(bareDir, "SKILL.md");
+  }
+  // .derived.md variant for memory refs
+  if (refType === "memory") {
+    const derivedPath = path.join(root, "memories", `${refName}.derived.md`);
+    if (fs.existsSync(derivedPath)) return derivedPath;
+  }
+  // Knowledge-specific: search subdirectories like knowledge/projects/, knowledge/tools/, etc.
+  if (refType === "knowledge") {
+    try {
+      const knowledgeDir = path.join(root, "knowledge");
+      if (fs.existsSync(knowledgeDir) && fs.statSync(knowledgeDir).isDirectory()) {
+        const entries = fs.readdirSync(knowledgeDir, { withFileTypes: true });
+        for (const entry of entries) {
+          if (!entry.isDirectory()) continue;
+          const subPath = path.join(knowledgeDir, entry.name, `${refName}.md`);
+          if (fs.existsSync(subPath)) return subPath;
+        }
+      }
+    } catch {
+      // Ignore errors reading directory
+    }
+  }
+  // Fallback: the refName may already encode the full stash-relative path
+  // (e.g. knowledge:skills/foo/references/bar where the file lives at
+  // <stash>/skills/foo/references/bar.md, not <stash>/knowledge/skills/...).
+  const directPath = path.join(root, `${refName}.md`);
+  if (fs.existsSync(directPath)) return directPath;
+  const directDir = path.join(root, refName);
+  if (fs.existsSync(directDir) && fs.existsSync(path.join(directDir, "SKILL.md"))) {
+    return path.join(directDir, "SKILL.md");
+  }
+  return null;
 }
 
 /**
--- a/src/commands/mv-cli.ts
+++ b/src/commands/mv-cli.ts
@@ -1,0 +1,557 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+/**
+ * `akm mv <ref> <new-name>` — rename an asset within its type directory
+ * (SPEC-7, stash-conventions-code-spec.md).
+ *
+ * The stash organization conventions' forced-rename procedure ("grep and fix
+ * inbound xrefs in the same pass") is agent-executable EXCEPT for the part
+ * only the CLI can do: a rename mints a new `entries` row (entry_key is
+ * UNIQUE), which orphans the utility_scores / utility_scores_scoped /
+ * embeddings / asset_salience rows keyed by entry_id — the "rename resets
+ * learned ranking" cost the conventions warn about. This verb does the whole
+ * pass: move the file, rewrite inbound refs, and re-key the index row IN
+ * PLACE so the accumulated history survives.
+ *
+ * Scope (v1, Experimental — see STABILITY.md):
+ *   - flat-markdown asset types only ({@link MV_SUPPORTED_TYPES});
+ *   - the primary writable stash only (no `--target`);
+ *   - the source ref must be the CANONICAL spelling — lint-resolver fallback
+ *     spellings are rejected with the canonical ref named (see
+ *     {@link resolveMoveSourcePath});
+ *   - wiki refs are rejected (wikis have their own xref + lint system);
+ *   - a memory's `.derived.md` twin moves together and keeps its
+ *     `entry_key === <base entry_key> + ".derived"` coupling; a twin ref
+ *     cannot be moved alone, and target names ending `.derived` are rejected
+ *     (reserved suffix).
+ *
+ * Ordering (partial-failure mitigation — the operation spans FS + DB and is
+ * NOT transactional): validate everything first, compute the full rewrite
+ * plan, apply the citer edits, rename the file(s) LAST among the FS steps,
+ * then re-key the index. The command is RE-RUNNABLE after an interruption:
+ * a crash before the rename leaves the source resolvable under its old ref
+ * (already-edited citers simply yield zero further rewrites on the retry);
+ * a crash after the rename but before the index re-key is healed by the next
+ * full `akm index` (at the cost of the utility history the re-key preserves).
+ * Graph tables (graph_files) key extractions by file path and stay stale
+ * until the next graph pass — acceptable, the graph is a derived cache.
+ */
+
+import fs from "node:fs";
+import path from "node:path";
+import { defineJsonCommand, output } from "../cli/shared";
+import { parseAssetRef, refToString } from "../core/asset/asset-ref";
+import { deriveCanonicalAssetNameFromStashRoot, TYPE_DIRS } from "../core/asset/asset-spec";
+import { type AkmAssetType, isWithin, resolveStashDir, toPosix } from "../core/common";
+import { UsageError } from "../core/errors";
+import { appendEvent } from "../core/events";
+import { getDbPath } from "../core/paths";
+import { warnVerbose } from "../core/warn";
+import { closeDatabase, openExistingDatabase, rebuildFts, rekeyEntryInPlace } from "../indexer/db/db";
+import { indexWrittenAssets, WRITE_PATH_INDEX_BUSY_TIMEOUT_MS } from "../indexer/index-written-assets";
+import { resolveSourceEntries } from "../indexer/search/search-source";
+import { refToRelPath, resolveRefPathInStash } from "./lint/base-linter";
+
+// ── Scope ─────────────────────────────────────────────────────────────────────
+
+/**
+ * Asset types `akm mv` can rename in v1: exactly the types whose canonical
+ * layout is one flat `.md` file per name (the `markdownSpec` family), so a
+ * rename is a single-file move and inbound refs are rewritable by complete-ref
+ * matching. Deliberately excluded:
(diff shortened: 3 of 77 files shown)


src/commands/read/knowledge.ts
```diff
@@ -181,18 +476,109 @@ export async function writeMarkdownAsset(options: {
...
+      const reason =
+        "its existing frontmatter is not a parseable YAML mapping — rewriting it would drop the values the parser could not read";
+      warn(
+        `Warning: superseded asset ${item.ref} was NOT demoted (${reason}). ` +
+          "The correction was written and cites it in xrefs; fix the old asset's frontmatter and re-run the correction with --force.",
+      );
+      superseded.push({ ref: item.ref, applied: false, reason });
+      continue;
+    }
+    writeSupersededEdge(item.filePath, result.ref);
```
fwdslsh-dev: 🔴 **#1 — `--supersedes` corrupts secret / env / task files.** `resolveSupersedesForWrite` places no type restriction, so `secret:`, `env:`, `task:`, and explicit-`.yaml` `workflow:` refs reach `writeSupersededEdge` → `mutateFrontmatter`, which **prepends a `---…---` YAML block over the raw file bytes** when the file has no existing frontmatter (the guard at :525 only fires when frontmatter already exists). `akm mv` excludes these types as "not markdown assets" — `--supersedes` does not.

Verified (exit 0, `applied:true`):
```sh
akm remember "key rotated" --supersedes secret:clientX/api-key
# secrets/clientX/api-key now starts with a "---\nsupersededBy: …\nbeliefState: superseded\n---" block
akm secret run secret:clientX/api-key KEY -- sh -c 'printf "%s" "$KEY"'  # injects the corrupted value
```
A task `.yml` becomes multi-document YAML that `yaml.parse` rejects.

**Fix:** gate `--supersedes` (and the demotion write) to markdown asset types — reuse mv's `MV_SUPPORTED_TYPES`, or require the resolved file to be `.md` — and reject `secret:`/`env:`/`task:`/`script:` with a `UsageError` before any write. Full detail + repro in the [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) (finding #1).

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,559 @@
...
+ *   - true — no index exists, the file was never indexed (nothing to
+ *     preserve; the next `akm index` picks it up fresh), or the row(s) were
+ *     re-keyed in place;
+ *   - false — an existing index could not be re-keyed (open/SQL error), or a
+ *     row for the moved file exists under some OTHER entry_key than the one
+ *     the canonical spelling derives (e.g. a differently-normalized stash
+ *     path at index time): that history is now stranded on a ghost row and
+ *     will NOT survive the next index run.
+ */
+function rekeyIndexForMove(opts: {
```
fwdslsh-dev, line 441: 🔴 **#2 — `akm mv` loses usage/utility history at the first full reindex.** `rekeyEntryInPlace` updates only the `entries` row; it never rewrites `usage_events.entry_ref`, which keeps the **old** ref — yet `mv` reports `utilityPreserved: true`. A plain incremental index may skip the dir (mtime preserved), so the loss is deferred, but `akm index --full` (which this PR's own `indexBodyOpening` docs tell users to run) detaches events to `entry_id = NULL` keyed by the old ref, `relinkUsageEvents → findEntryIdByRef('…/old-note')` matches nothing, and utility resets to default — the exact reset `mv` exists to prevent. (The incremental dir-stale path hard-deletes the events by `entry_id`, db.ts:601-605.)

**Fix:** in `rekeyEntryInPlace`, also `UPDATE usage_events SET entry_ref = ? WHERE entry_ref = ?` (old→new canonical ref) for the base **and** the `.derived` twin. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #2.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,559 @@
...
+ * layout is one flat `.md` file per name (the `markdownSpec` family), so a
+ * rename is a single-file move and inbound refs are rewritable by complete-ref
+ * matching. Deliberately excluded:
+ *   - `wiki` — wikis carry their own xref + lint system (`akm wiki lint`);
+ *   - `skill` — the canonical layout is a multi-file `skills/<name>/SKILL.md`
+ *     directory (a directory rename, out of v1 scope);
+ *   - `script` — unresolvable by the slug resolver (contract-pinned);
+ *   - `task` / `env` / `secret` — not markdown assets.
+ */
+const MV_SUPPORTED_TYPES: readonly string[] = [
```
fwdslsh-dev: 🔴 **#3 — `akm mv` corrupts or cannot move YAML workflows.** `workflow` is in `MV_SUPPORTED_TYPES`, but `refToRelPath` → `workflowSpec.toAssetPath` probes `fs.existsSync` against a **cwd-relative** path (`TYPE_DIRS.workflow` is the bare `"workflows"`, not stash-rooted) and falls back to `<name>.md`. With `workflows/deploy.yaml` on disk:
- cwd = stash root → `akm mv workflow:deploy release` exits 0 and `renameSync` produces `workflows/release.md` **still containing YAML** (silent corruption).
- cwd elsewhere → exit 2 `"Cannot resolve workflow:deploy"` though the asset exists.

Both verified end-to-end. An unrelated `workflows/deploy.*` in the user's cwd also steers which path `mv` operates on.

**Fix:** for v1, **drop `workflow` from `MV_SUPPORTED_TYPES`** (the docstring already claims "flat `.md` only"), or resolve old/new paths against the **stash dir** and preserve the source file's real extension instead of forcing `.md`. Same cwd-relative resolution also drives #7. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #3.

---

src/indexer/db/db.ts
```diff
@@ -418,6 +418,87 @@ export function getBaseBeliefStatesForDerivedTwins(db: Database, twinIds: number
...
+
+  db.transaction(() => {
+    const stale = db.prepare("SELECT id FROM entries WHERE entry_key = ?").get(opts.newEntryKey) as
+      | { id: number }
+      | undefined
+      | null;
+    if (stale && stale.id !== row.id) {
+      db.prepare("DELETE FROM entries_fts WHERE entry_id = ?").run(stale.id);
+      db.prepare("DELETE FROM entries_fts_dirty WHERE entry_id = ?").run(stale.id);
+      db.prepare("DELETE FROM entries WHERE id = ?").run(stale.id);
```
fwdslsh-dev, line 521: 🔴 **#4 — stale-row eviction hits a FK constraint and silently rolls back the whole re-key.** When a rename targets a name whose previous file was deleted but whose index row lingers, this bare `DELETE FROM entries` cleans only `entries`/`entries_fts`/`entries_fts_dirty` — unlike `deleteRelatedRows` (which also cleans `embeddings`, `entries_vec`, `utility_scores(_scoped)`, `usage_events`). The mv connection runs with `PRAGMA foreign_keys = ON`, and `embeddings` has a **non-CASCADE** FK to `entries(id)`.

Verified in bun:sqlite: with an `embeddings` child row present, this `DELETE` throws `FOREIGN KEY constraint failed` and **rolls back the entire re-key transaction** (row update, twin re-key, FTS rebuild). `mv-cli.ts:292` reduces it to a verbose-only warning + `utilityPreserved:false`, so the asset keeps its old `entry_key` and its history resets on the next index. In the no-embeddings case, `utility_scores_scoped`/`usage_events` orphan permanently.

**Fix:** route the displaced-row eviction through `deleteRelatedRows`/`deleteEntriesByIds` so child rows are cleaned before the `entries` delete; and don't swallow a re-key failure to `warnVerbose` while exiting 0. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #4.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,559 @@
...
+  type: string;
+  oldName: string;
+  newName: string;
+  oldPath: string;
+  newPath: string;
+  toRef: string;
+  twinOldPath: string | null;
+  twinNewPath: string | null;
+}): boolean {
+  const dbPath = getDbPath();
```
fwdslsh-dev, line 453: 🔴 **#5 — `akm mv` never re-keys `state.db` salience / outcome rows.** `rekeyIndexForMove` opens only `getDbPath()` (index.db). `asset_salience` and `asset_outcome` live in **state.db** and are keyed by `asset_ref TEXT` — **not** `entry_id` as the module docstring (line 13) and `db.ts:442-444` claim. So they stay keyed to the old ref. `loadSalienceRankScores` looks rows up by `makeAssetRef(type, CURRENT name)`, so after a rename the salience boost (up to ~1.18×) is lost until the next improve run re-mints a row — and a distill-written content-derived `encoding_salience` never carries over. `asset_outcome` history is likewise stranded; orphan old-ref rows linger. The help text (mv-cli.ts:308-311) promises "salience … survives the rename."

**Fix:** open state.db and `UPDATE asset_salience SET asset_ref = ? WHERE asset_ref = ?` (and `asset_outcome`), old→new `makeAssetRef`, handling a pre-existing row at the new ref. Correct the docstrings. If deferring, drop the "salience survives" claim. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #5.

---

src/commands/read/knowledge.ts
```diff
@@ -129,6 +137,277 @@ export async function readKnowledgeInput(
...
+  const xrefs: string[] = [];
+  for (const raw of rawXrefs) {
+    const trimmed = raw.trim();
+    if (trimmed && !xrefs.includes(trimmed)) xrefs.push(trimmed);
+  }
+  if (xrefs.length === 0) return xrefs;
+
+  const cfg = loadConfig();
+  const stashRoot = resolveWriteTarget(cfg, target).source.path;
+  const extraStashRoots = resolveSourceEntries(stashRoot, cfg)
```
fwdslsh-dev: 🟠 **#6 — `--xref` can't resolve working-stash refs when writing to a named target.** The root set here is `[writeTarget, ...resolveSourceEntries(writeTarget)]`. Passing the write-target path as `overrideStashDir` bypasses `resolveStashDir()`, and `config.sources` has no synthesized entry for a top-level `stashDir`, so the **primary working stash is absent** whenever `--target`/`defaultWriteTarget` routes to a named source. The sibling `resolveSupersedesForWrite` (:330-337) explicitly adds `resolveStashDir({readOnly:true})` — so the same ref resolves via `--supersedes` but not `--xref`. (The docstring's "mirrors lint's root set" is false; lint roots at `resolveStashDir`.)

Repro: top-level `stashDir ~/stash` (not in `sources[]`) + source `team` + `defaultWriteTarget team`, `~/stash/memories/local-note.md` exists → `akm remember "note" --xref memory:local-note` → `UsageError "--xref ref did not resolve"`.

**Fix:** add `resolveStashDir({readOnly:true})` to the `--xref` root set (mirror `--supersedes`); ideally extract one shared root-set helper used by both + lint. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #6.

---

src/commands/read/knowledge.ts
```diff
@@ -129,6 +137,277 @@ export async function readKnowledgeInput(
...
+    .map((s) => s.path)
+    .filter((p) => p !== stashRoot && fs.existsSync(p));
+  const allRoots = [stashRoot, ...extraStashRoots];
+
+  const unresolved: string[] = [];
+  for (const ref of xrefs) {
+    const colonIdx = ref.indexOf(":");
+    const refType = colonIdx > 0 ? ref.slice(0, colonIdx) : "";
+    const refName = colonIdx > 0 ? ref.slice(colonIdx + 1) : "";
+    const relPath = refType && refName ? refToRelPath(refType, refName) : null;
```
fwdslsh-dev: 🟠 **#7 — write-time validation is stricter than lint: `script:` always rejected, `workflow:` cwd-dependent.** These resolvers treat `refToRelPath === null` as "unresolved → abort exit 2", inverting lint's deliberate fail-open (`checkMissingRefs` does `if (relPath === null) continue`, base-linter.ts:323).

- `script:` is contract-pinned to return `null` (base-linter.ts:177), so **no spelling of a `script:` xref can ever pass** — yet this PR's own `script.md`/`domains.md`/`backlinks.md` tell agents to write `--xref script:build/release`, and the error hint (`akm search --type script`) *finds* the asset. Verified: `akm remember "tip" --xref script:build/release` exits 2 for an existing script.
- `workflow:` validation is cwd-dependent (same `toAssetPath` cwd probe as #3): `--xref workflow:deploy` (only `deploy.yaml` on disk) exits 2 from any cwd but the stash root.

**Fix:** for null-relPath types, mirror lint's fail-open (skip existence validation); resolve `workflow:` against the stash dir (shared with #3); parse refs with `parseAssetRef` instead of `indexOf(":")` so origin-prefixed refs get a structured message. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #7.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,559 @@
...
+      );
+    }
+
+    // Memory `.derived.md` twin: moves together with its base (the entry_key
+    // suffix coupling the belief-state inheritance relies on).
+    const isBaseMemory = source.type === "memory" && !source.name.endsWith(".derived");
+    const twinOldPath = isBaseMemory ? oldPath.replace(/\.md$/, ".derived.md") : null;
+    const hasTwin = twinOldPath !== null && fs.existsSync(twinOldPath);
+    const twinNewPath = hasTwin ? newPath.replace(/\.md$/, ".derived.md") : null;
+    if (twinNewPath && fs.existsSync(twinNewPath)) {
```
fwdslsh-dev: 🟠 **#8 — `akm mv` adopts an orphaned `.derived.md` twin.** The target twin-collision check is gated on the **source** memory having a twin (`twinNewPath` is null otherwise), so renaming a twin-less memory onto a name whose orphaned `memories/<name>.derived.md` already exists passes the base-only `fs.existsSync(newPath)` check (:440) and silently adopts the stranger file. Coupling is **query-time** (the orphan's row already has `derived_from='memory:<name>'`), so search-hit enrichment immediately swaps the renamed memory's description/searchHints/tags for the stale distillation, and the orphan inherits its beliefState via the `entry_key + '.derived'` join. Orphan twins are produced by consolidate/dedup deletes (single-file unlink, no twin cleanup).

**Fix:** always check `fs.existsSync(newPath.replace(/\.md$/, ".derived.md"))` for a target collision (unconditionally, not only when the source has a twin) and reject. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #8.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,559 @@
...
+    } catch {
+      return;
+    }
+    for (const entry of entries) {
+      if (entry.name.startsWith(".")) continue;
+      const full = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        if (entry.name === "registry") continue;
+        walk(full);
+      } else if (entry.isFile() && entry.name.endsWith(".md")) {
```
fwdslsh-dev: 🟠 **#9 — `akm mv` leaves task-`.yml` citers dangling.** `collectMarkdownFiles` filters `endsWith(".md")`, but lint scans `tasks/*.yml` through the same missing-ref body check (lint/index.ts:120, `body === raw`), and task YAML legitimately carries refs (`workflow: workflow:…`, `prompt: agent:/memory:…`, src/tasks/parser.ts:13). Task citers of a moved ref are neither rewritten nor reported in `readOnlyCiters`.

Repro: `tasks/nightly.yml` with `workflow: workflow:daily-backup`; `akm mv workflow:daily-backup backup/daily` reports `rewrote:[]`/`readOnlyCiters:[]`; the scheduled task now dangles and the next `akm lint` flags it.

**Fix:** extend the citer scan to task YAML (`tasks/*.yml`) — ideally share lint's file collection so the two don't diverge (cleanup C-walker) — and rewrite refs there (or at least report them in `readOnlyCiters`). See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #9.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,559 @@
...
+
+/**
+ * Boundary grammar shared with lint's `REF_RE` (base-linter.ts): a ref starts
+ * at line start or after whitespace / backtick / quote / `(`, and its slug
+ * runs until whitespace, `"`, `'`, backtick, `)`, `]`, `>`, or `,`. Complete-
+ * ref matching is what keeps a longer ref sharing the old ref as a prefix
+ * (e.g. `memory:a/base-note-extra` when moving `memory:a/base-note`)
+ * untouched.
+ */
+const REF_PREFIX_SRC = "(^|[\\s`\"'(])";
```
fwdslsh-dev: 🟠 **#10 — `akm mv` skips flow-style YAML list and bracketed refs.** `REF_PREFIX_SRC`'s boundary class omits `[`, so a ref preceded by `[` never matches `buildRefPattern`. Verified end-to-end:
- `xrefs: [memory:projectA/old-note]` (flow style) is left byte-identical after `mv` (exit 0), then flagged missing-ref by **this PR's own new YAML-parsing lint pass** post-move.
- `xrefs: [a, memory:old]` gets only the **second** element rewritten — one list partially rewritten in a single run.
- The body form `see [memory:old]` dangles with **no lint flag at all** (lint's `REF_RE` shares the `[`-less prefix class) — completely silent.

**Fix:** add `[` to `REF_PREFIX_SRC`. Since this grammar is meant to match lint's `REF_RE`, export the boundary/slug fragments from `base-linter.ts` and import them here (cleanup C-grammar); applying the same fix to `REF_RE` also closes the silent body-bracket gap. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #10.
*(Related, lower sev: refs followed by sentence punctuation `…old-note.` are also not rewritten, but lint's shared grammar flags those pre+post move — user-visible, not silent.)*

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,559 @@
...
+}
+
+/**
+ * Build the rewrite pattern for one old ref. When the source is a base
+ * memory, the optional `.derived` tail also rewrites explicit twin refs
+ * (`memory:a/note.derived` → `memory:a/new.derived`) — the twin file moves
+ * together, so its ref must too. The tail group is empty-capturing otherwise
+ * so the replacer's group indices stay stable.
+ */
+function buildRefPattern(fromRef: string, includeDerivedTail: boolean): RegExp {
```
fwdslsh-dev: 🟠 **#11 — `akm mv` leaves alias-spelling citers dangling.** `buildRefPattern` matches only the canonical extensionless spelling, but the resolver stack treats several alternates as the same asset. Verified: citers using `knowledge:old-guide.md` (`.md` suffix accepted by `markdownSpec.toAssetPath`; lint-clean pre-move), the knowledge-subdir alias `knowledge:guide` (for `knowledge/guides/guide.md`), and `local//memory:…` (produced by `makeAssetRef`, accepted by `parseAssetRef`/`akm show`) are all left untouched and appear in neither `rewrote` nor `readOnlyCiters`. The `.md`/subdir aliases are flagged by lint post-move; the `local//` form is invisible to lint's `REF_RE`, so it breaks with no diagnostic ever.

**Fix:** prefer a grammar-based rewrite — for each candidate file, resolve every ref-shaped token via the shared resolver and rewrite any that resolves to the moved file (covers `.md`, subdir, and `local//` spellings). At minimum, generate patterns for the known alias forms and report anything left unresolved. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #11.

---

src/commands/sources/stash-cli.ts
```diff
@@ -197,21 +203,49 @@ export const importKnowledgeCommand = defineJsonCommand({
...
+    }
     const stashDir = resolveWriteTarget(loadConfig(), args.target).source.path;
     const { content, preferredName } = await readKnowledgeInput(args.source, { stashDir });
+    // Imported docs may carry their own frontmatter: merge (dedupe-append)
+    // BEFORE the write so write-path indexing sees the final content and no
+    // second frontmatter block is ever nested.
     const result = await writeMarkdownAsset({
       type: "knowledge",
-      content,
+      content: mergeXrefsIntoContent(content, xrefs),
```
fwdslsh-dev, line 252: 🟠 **#12 — `--xref`/`--supersedes` without `--name` produce a random slug.** These flags force the *structured* write path, whose `writeMarkdownAsset` call omits the `preferredName` guard the hot path uses. `mergeXrefsIntoContent` here prepends the `---` frontmatter block **before** name inference, so `inferAssetName` takes `---` as the first non-empty line, `slugify` strips it to `""`, and the random `<type>-<epoch>-<rand>` fallback fires. Same root cause in `remember-cli.ts:295` (that line sits between diff hunks, so no inline anchor there). Verified:
```sh
akm import - --xref knowledge:auth-flow < notes.md   # → knowledge:knowledge-<ts>-<rand>  (without --xref → knowledge:auth-notes)
akm remember "…rollout" --supersedes memory:old-note  # → memory:memory-<ts>-<rand>
```

**Fix (import):** infer the name from `content` **before** `mergeXrefsIntoContent`. **Fix (remember-cli.ts:295):** pass `preferredName: inferAssetName(body, "memory")` on the structured `writeMarkdownAsset` call, matching the hot path. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #12.

---

src/commands/read/knowledge.ts
```diff
@@ -181,18 +476,109 @@ export async function writeMarkdownAsset(options: {
...
+    }
+    writeSupersededEdge(item.filePath, result.ref);
+    superseded.push({ ref: item.ref, applied: true });
+    const files = demotedByRoot.get(item.stashRoot) ?? [];
+    files.push(item.filePath);
+    demotedByRoot.set(item.stashRoot, files);
+  }
   // 0.9.0 (issue #507): single batch commit at the write boundary for git
   // targets. No-op for filesystem/primary-stash targets.
   commitWriteTargetBoundary(target, `Update ${formatRefForMessage(ref)}`);
```
fwdslsh-dev, line 700: 🟠 **#13 — a demotion fs error leaves the new asset written but unindexed and uncommitted.** The correction file is written first (:495); `writeSupersededEdge` (:535) then runs with no `try/catch` (the only catch, :520-524, deliberately re-throws). Any fs/YAML error demoting the **old** asset (read-only file, concurrent delete) aborts before `commitWriteTargetBoundary` and `indexWrittenAssets` — so the new asset is on disk but **absent from the index** (breaking "searchable immediately") and, on a git target, uncommitted. The module's own `superseded:[{applied:false}]` partial-failure report is bypassed, and a naive re-run exits 2 `RESOURCE_ALREADY_EXISTS` until `--force`.

**Fix:** wrap `writeSupersededEdge` in `try/catch`; on failure still commit + index the new asset and report `superseded:[{applied:false, error}]` (reuse the existing shape). A demotion error must not abort the correction write. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #13.

---

src/commands/improve/memory/memory-belief.ts
```diff
@@ -71,3 +71,39 @@ export function writeContradictEdge(filePath: string, contradictedByRef: string)
...
+ * Idempotent: if `supersededByRef` is already in `supersededBy` and the file
+ * is already demoted, the file is not rewritten. Multiple corrections
+ * sorted-set-append their refs.
+ *
+ * @param filePath        - Absolute path to the asset markdown file.
+ * @param supersededByRef - The ref of the correction that supersedes this asset.
+ */
+export function writeSupersededEdge(filePath: string, supersededByRef: string): void {
+  mutateFrontmatter(filePath, (parsed) => {
+    const existing: string[] = Array.isArray(parsed.data.supersededBy) ? (parsed.data.supersededBy as string[]) : [];
```
fwdslsh-dev: 🟠 **#14 — a scalar `supersededBy` edge is silently destroyed.** This line reads `Array.isArray(parsed.data.supersededBy) ? … : []`, so a pre-existing **scalar-string** `supersededBy: memory:first-fix` is treated as empty and replaced by only the new ref — the old edge is dropped with no warning. Scalar values are live data elsewhere: the indexer's `normalizeNonEmptyStringList` and `memory-improve`'s `refArray` both accept a scalar, and lint never flags it (#15a). The sibling `mergeXrefsIntoContent` in this same PR explicitly promotes a scalar `xrefs` to a list — inconsistent.

**Fix:** mirror `mergeXrefsIntoContent` — treat a scalar string as `[value]` before merging. Apply the same to the pre-existing `writeContradictEdge` (same bug, now reachable via `--supersedes`). See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #14.

---

src/commands/lint/base-linter.ts
```diff
@@ -617,6 +655,35 @@ export abstract class BaseLinter implements AssetLinter {
...
+      // Non-ref-shaped values (URLs, `raw/<slug>`, `<placeholder>`
+      // templates, shell vars) fall out via checkMissingRefs' guards.
+      //
+      // Skipped when ctx.frontmatter is null: on the task/YAML path
+      // (lint/index.ts) the whole file IS the body (`body === raw`), so
+      // ref values under these keys are already caught by the body scan
+      // above — running the pass again would double-report each dangling
+      // ref. Md files without a frontmatter block also land here, with
+      // empty ctx.data, so nothing is lost for them either.
+      if (ctx.frontmatter !== null) {
```
fwdslsh-dev: 🟠 **#15 — the new lint xref pass has two escape hatches.**
- **(a) scalar values:** `readRefsArray` returns `null` for any non-array and the loop `continue`s, so a scalar `supersededBy: memory:gone` (or scalar `xrefs`/`contradictedBy`) is never validated, and the body scan can't see frontmatter. Realistic because the indexer honors scalar edges (see #14).
- **(b) YAML-path task files:** this pass is gated on `ctx.frontmatter !== null`, which is null on the task/YAML path — and there **any authoritative `refs:` array (even a valid non-empty one) suppresses the body scan**. So `tasks/t1.yml` with `refs: [memory:exists]` + `xrefs: [memory:gone]` reports **zero** missing-ref findings. Verified end-to-end; violates this PR's own SPEC-1 test-plan item ("`refs: []` does NOT suppress the xrefs scan").

**Fix:** (a) validate scalars too — use a reader that accepts both a scalar string and an array. (b) change the gate to `if (ctx.frontmatter !== null || explicitRefs !== null)` — no double-report risk, since when `explicitRefs !== null` the body scan doesn't run. See [review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md) #15.

---

fwdslsh-dev (commented):
## xhigh code review — 15 verified findings (5 data-loss / corruption)

Reviewed `3ca2779`. Method: 10 finder angles → 60 raw candidates → 33 deduped → one adversarial verifier each, many with **end-to-end repros against the real CLI**. 2 candidates refuted.

The lint (SPEC-1), search ref-prefix (SPEC-4), category capture (SPEC-6), and body-opening (SPEC-8) work is largely sound. **The serious problems cluster in the two new write paths — `--supersedes` (SPEC-5) and `akm mv` (SPEC-7).** `akm mv`'s headline promise ("inbound refs rewritten, ranking history preserved") is not met along several independent paths, and `--supersedes` can silently corrupt raw non-markdown assets.

Each finding is an **inline comment** on its line with a repro and a concrete fix. The full report — repros, code, and fix direction for every item — is committed alongside as a durable artifact:

📄 **[code-review/pr-715-stash-conventions-review.md](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review.md)**

### 🔴 Data loss / corruption
| # | Where | Issue |
|---|---|---|
| 1 | `knowledge.ts:535` | `--supersedes` prepends a frontmatter block over raw **secret / env / task** files (verified: `akm secret run` then injects corrupted bytes) |
| 2 | `mv-cli.ts:234` | `akm mv` never rewrites `usage_events.entry_ref` → utility history lost at first `akm index --full`, despite `utilityPreserved:true` |
| 3 | `mv-cli.ts:70` | `akm mv` of a `.yaml` workflow renames it to `.md` (corrupt) or fails to resolve — cwd-dependent |
| 4 | `db.ts:488` | stale-row eviction hits a non-CASCADE FK → whole re-key silently rolls back |
| 5 | `mv-cli.ts:245` | `akm mv` never re-keys `state.db` salience/outcome rows (docstring falsely says `entry_id`-keyed) |

### 🟠 Silent incorrect behavior
| # | Where | Issue |
|---|---|---|
| 6 | `knowledge.ts:175` | `--xref` can't resolve working-stash refs when writing to a named `--target` (asymmetric with `--supersedes`) |
| 7 | `knowledge.ts:185` | write-time validation rejects valid `script:` refs always, `workflow:` cwd-dependently (stricter than lint) |
| 8 | `mv-cli.ts:454` | `akm mv` adopts an orphaned `.derived.md` twin (target twin-collision check gated on source having a twin) |
| 9 | `mv-cli.ts:142` | `akm mv` ignores task-`.yml` citers (lint scans them) → dangling scheduled-task refs |
| 10 | `mv-cli.ts:91` | `akm mv` skips flow-style `xrefs: [ref]` and bracketed body refs (boundary class omits `[`) |
| 11 | `mv-cli.ts:105` | `akm mv` skips `.md`-suffix / subdir-alias / `local//` citer spellings the resolver accepts |
| 12 | `stash-cli.ts:241`, `remember-cli.ts:295` | `--xref`/`--supersedes` without `--name` → random `<type>-<epoch>-<rand>` slug |
| 13 | `knowledge.ts:543` | a demotion fs error leaves the new correction written but unindexed + uncommitted, no partial report |
| 14 | `memory-belief.ts:97` | a scalar `supersededBy` edge is silently dropped when demoting |
| 15 | `base-linter.ts:672` | the new lint xref pass misses scalar values and task-`.yml` files carrying an authoritative `refs:` array |

### 🟡 Below the cut (verified; lower severity)
- `metadata.ts:1184` — `extractBodyOpening` captures leading **HTML comments** as prose (incl. this PR's own skeleton convention facts) when `indexBodyOpening` is on.
- `index-written-assets.ts:79` — write-path indexer silently drops `workflow` entries (comment now outdated); a moved workflow / workflow citer keeps stale `search_text`.
- `metadata.ts:1192` — `slice(0, 279)` can split a surrogate pair into FTS/embedding text.
- `memory-belief.ts:106` — demoting a `contradicted` asset to `superseded` *weakens* its demotion and leaves `contradictedBy` inconsistent.

### ⚪ Cleanups that also prevent regressions of the above
Share one `.md`/`.yml` walker with lint (→ #9); export lint's `REF_RE` grammar and import it in `mv` (→ #10); use `parseAssetRef` instead of `indexOf(":")` (→ #7); one shared xref/supersedes root-set helper (→ #6); unify the 4 belief-state enumerations; dedupe `resolveSupersedesForWrite`/`resolveXrefsForWrite` (also removes 3–4× redundant `loadConfig` per write). Details in the doc.

### Refuted (investigated, not real)
- ref-prefix `--type` "degradation" is spec-pinned intended behavior (CHANGELOG + `search-ref-prefix.test.ts:195`) — at most a doc-hint polish.
- lint ref-resolution memoization — negligible marginal cost over the pre-existing body scan.

**Suggested triage:** #1 first (clearest data-loss), then the mv re-key/history cluster #2/#4/#5, then #3; #6–#15 are correctness fixes; the ⚪ walker/grammar/refparse cleanups are worth doing alongside #9/#10/#7.

*Automated review — filed as comments, not a formal block; the maintainer decides what's in scope for this PR.*

---

itlackey:
All 15 findings plus the 4 below-the-cut items were independently re-verified against the code — **every one reproduced as claimed, none disputed** — and are now fixed on this branch, each behind a regression test that failed pre-fix. Thanks for an unusually precise review.

**Fix commits, mapped to findings:**

| Commit | Findings | Highlights |
|---|---|---|
| `3e2db4d` (+fmt `9a…`) | below-cut ×3 | `extractBodyOpening` skips HTML comments + never splits a surrogate pair; write-path indexer now upserts workflow entries *with* their `workflow_documents` row |
| `450201e` | #15a, #15b, #10-grammar | scalar frontmatter edges validated; task-`.yml` `refs:` no longer suppresses the xref pass; `[` added to the boundary class; grammar fragments exported so `mv` and lint share one grammar |
| `dbff7e9` | #1, #6, #7, #12, #13, #14, below-cut | **#1**: `--supersedes` refuses raw types (secret/env/task/script) and any non-`.md` target before any write; shared `resolveWriteRefRoots()` for both flags (#6); lint-mirroring fail-open for `script:` + stash-rooted `workflow:` resolution (#7); stable slugs under `--xref`/`--supersedes` (#12); demotion failure no longer strands the correction unindexed (#13); scalar edges preserved in both belief writers (#14); superseding a `contradicted` asset keeps the stronger state |
| `6a1d844` | #2, #3, #4, #5, #8, #9, #10, #11 | `usage_events.entry_ref` re-keyed in-transaction (base + `.derived` twin), survives `akm index --full` (#2); displaced-row eviction via `deleteRelatedRows` — no more FK rollback, plus an additive `warnings` report field (#4); `asset_salience`/`asset_outcome` re-keyed in state.db and the false `entry_id` docstrings corrected (#5); `workflow` dropped from `MV_SUPPORTED_TYPES` with a clear error (#3); unconditional target-twin collision check (#8); task-`.yml` citers rewritten (#9); flow-style/bracketed refs via the shared grammar (#10); `.md`-suffix, `local//`, and subdir-alias spellings rewritten via patterns + a resolver-backed second pass (#11) |

The ⚪ cleanups that prevent regressions of the fixed items landed alongside them (shared grammar export, shared root-set helper, `parseAssetRef` in write validation). The broader refactors (belief-state enum unification, resolver dedupe beyond the shared helpers) are left for a follow-up.

Full gate is green over the combined stack: `bun run check` exits 0 — 26,248 unit / 765 integration pass, contracts (incl. ref-resolver and §9.4 CLI surface) and all repo lint guards clean.

---

itlackey:
@codex do a full code review for this PR

src/indexer/db/db.ts
```diff
@@ -418,6 +418,131 @@ export function getBaseBeliefStatesForDerivedTwins(db: Database, twinIds: number
...
+    ).run(opts.newEntryKey, path.dirname(opts.newFilePath), opts.newFilePath, entryJson, searchText, row.id);
+    if (opts.newDerivedFrom !== undefined) {
+      db.prepare("UPDATE entries SET derived_from = ? WHERE id = ?").run(opts.newDerivedFrom, row.id);
+    }
+    // Re-point usage history at the new ref (see the docstring): the bare
+    // spelling exactly, and the origin-qualified spelling by rewriting only
+    // the part after the last `//` (stored origins never contain `//`, so a
+    // qualified ref has exactly one — same normalization as
+    // getRetrievalCounts). Legacy DBs may predate usage_events.
+    bestEffort(() => {
```
fwdslsh-dev, line 534: 🟠 **R2-1 (new defect in the #2 fix) — `mv` adopts a deleted stranger's usage history at the target ref.** This `usage_events` block only rewrites old→new. Detached orphan events (`entry_id NULL`) already sitting **at the new ref** are never evicted: no stale `entries` row exists, so the eviction path (:511-518) never runs, and `deleteRelatedRows` only deletes by `entry_id` anyway.

Repro (e2e at HEAD): `memories/dead.md` accumulates 10 events; file deleted; `akm index --full` detaches them (`entry_ref 'memory:dead'`, retained 90 days). `akm mv memory:live dead` → exit 0, `utilityPreserved:true`, no warnings. `getRetrievalCounts('memory:dead')` returns 5 immediately, and the next `--full` relinks all 10 stranger events by `entry_ref` (db.ts:1845) — verified utility 0.30 / showCount 5 for an asset with zero real usage.

**Fix:** inside the rekey transaction, `DELETE FROM usage_events WHERE entry_ref IN (<new-ref spellings>) AND entry_id IS NULL` **before** the old→new UPDATE — mirroring the "live asset's history wins" collision policy this same commit implements for state.db rows (mv-cli.ts:435-436) and stale entries rows. Full detail: [round-2 review doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md) R2-1.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,778 @@
...
+        try {
+          db.transaction(() => {
+            for (const [oldRef, newRef] of pairs) {
+              const moved = db.prepare(`SELECT asset_ref FROM ${table} WHERE asset_ref = ?`).get(oldRef);
+              if (!moved) continue;
+              db.prepare(`DELETE FROM ${table} WHERE asset_ref = ?`).run(newRef);
+              db.prepare(`UPDATE ${table} SET asset_ref = ? WHERE asset_ref = ?`).run(newRef, oldRef);
+            }
+          })();
+        } catch {
```
fwdslsh-dev: 🟠 **R2-2 (incomplete fix of #5) — state.db re-key failures are silently swallowed.** This per-table bare `catch {}` (commented "Table missing on an older state.db") wraps the **entire** re-key transaction, so every SQLite error — including `SQLITE_BUSY` past the 5s busy timeout — is eaten. Only open/PRAGMA failures reach the outer catch that feeds the report's `warnings` field, contradicting the docstring's "any error reduces to a warning in the report" (:440-442).

Repro (e2e at HEAD): seeded `asset_salience` row for `memory:sal`; a second connection held `BEGIN IMMEDIATE` for 25s; `akm mv memory:sal sal2` → exit 0, `utilityPreserved:true`, **no `warnings` key, empty stderr** — and `asset_salience` still keyed `'memory:sal'`. Control without the lock re-keys correctly. Per the docstring, the next improve run re-mints a type-stub row at the new ref, permanently losing a content-derived salience — silently.

**Fix:** discriminate in the inner catch — match `no such table` (or probe `sqlite_master`) for the legacy-schema case; route everything else to the `stateWarning` path. See [round-2 doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md) R2-2.

---

src/commands/sources/stash-cli.ts
```diff
@@ -197,21 +204,54 @@ export const importKnowledgeCommand = defineJsonCommand({
...
+    // import (no filename-derived preferredName) therefore pre-infers the name
+    // from the pre-merge content so --xref/--supersedes never change the slug.
     const result = await writeMarkdownAsset({
       type: "knowledge",
-      content,
+      content: mergeXrefsIntoContent(content, xrefs),
       name: args.name ?? (isHttpUrl(args.source) ? preferredName : undefined),
       fallbackPrefix: "knowledge",
-      preferredName,
+      preferredName: preferredName ?? (xrefs.length > 0 ? inferAssetName(content, "knowledge") : undefined),
```
fwdslsh-dev: 🟡 **R2-3 (incomplete fix of #12) — stdin doc carrying its own frontmatter + `--xref` still gets a random slug.** The fix infers the name from raw **pre-merge** content, but when the piped doc already has frontmatter its first non-empty line is the `---` fence, so the random `knowledge-<epoch>-<rand>` slug still fires.

Repro (e2e at HEAD): `printf -- '---\ndescription: x\n---\n\n# Frontmattered Guide\n\nBody.\n' | akm import - --xref knowledge:legacy-guide` → `knowledge:knowledge-1783840417073-5w15r`. The new integration test (`import-stdin-xref-slug.test.ts`) only pipes a frontmatter-less body, so its "same slug with and without --xref" contract isn't exercised for exactly the doc class `mergeXrefsIntoContent`'s merge branch exists for. (Mitigating: the plain no-flag run also random-slugs this input — pre-existing on main — so this is a half-applied fix, not a flag-caused regression.)

**Fix (one-liner, verified):** `inferAssetName(parseFrontmatter(content).content, "knowledge")` → yields `frontmattered-guide`. Add a frontmattered-stdin case to the test. See [round-2 doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md) R2-3.

---

src/commands/improve/memory/memory-belief.ts
```diff
@@ -58,9 +78,7 @@ export type {
  */
 export function writeContradictEdge(filePath: string, contradictedByRef: string): void {
   mutateFrontmatter(filePath, (parsed) => {
-    const existing: string[] = Array.isArray(parsed.data.contradictedBy)
-      ? (parsed.data.contradictedBy as string[])
-      : [];
+    const existing = readEdgeList(parsed.data.contradictedBy);
     if (existing.includes(contradictedByRef)) return null; // Already written — idempotent.
```
fwdslsh-dev: 🟠 **R2-4 (regression from the #14 fix) — `writeContradictEdge` no longer repairs a missing demotion.** The scalar promotion makes this guard fire for scalar `contradictedBy` values, but unlike its sibling the guard **ignores `beliefState`**: a file with scalar `contradictedBy: memory:newer` and no beliefState is now a permanent no-op. At base `3ca2779` the same call rewrote the file and set `beliefState: contradicted` (scalars read as `[]` — zero data loss in this same-ref case). The same commit made `writeSupersededEdge`'s guard state-aware (`existing.includes(ref) && currentState === nextState`, :125); the asymmetry looks like an oversight.

Impact: nothing downstream derives demotion from the edge alone (`beliefStateBoost` requires the field; `--belief current` passes `undefined`), and `handleContradictOp` (consolidate.ts:2663) increments `counts.contradicted++` on the no-op, reporting success — the memory keeps full rank forever. Verified empirically against both HEAD and base behavior.

**Fix:** mirror the sibling: `const nextState = currentState === "archived" ? currentState : "contradicted"; if (existing.includes(contradictedByRef) && currentState === nextState) return null;` See [round-2 doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md) R2-4.

---

src/commands/lint/base-linter.ts
```diff
@@ -150,12 +150,33 @@ function buildRefTypeAlternation(): string {
...
+ *
+ * `REF_BOUNDARY_PREFIX_CLASS_SRC` is the character class a ref may start
+ * after (a ref also matches at line start): whitespace, backtick, quote,
+ * `(`, or `[` — the `[` admits markdown-link-style refs like
+ * `see [memory:foo]`, which the legacy class silently skipped.
+ *
+ * `REF_SLUG_CHAR_CLASS_SRC` is the character class a ref's `<type>:<slug>`
+ * token is made of; the first excluded character ends the ref.
+ */
+export const REF_BOUNDARY_PREFIX_CLASS_SRC = "[\\s`\"'(\\[]";
```
fwdslsh-dev: 🟡 **R2-5 (residual of #10/#15b) — refs after a bare comma still invisible to lint AND mv.** `,` terminates a slug (`REF_SLUG_CHAR_CLASS_SRC` excludes it) but is **not** in `REF_BOUNDARY_PREFIX_CLASS_SRC`, so a ref immediately after a bare comma — the no-space flow list `xrefs: [memory:a,memory:b]`, valid YAML — can never start a match in lint's `REF_RE` or mv's shared rewrite grammar.

Repro (e2e at HEAD): task `.yml` **without** a `refs:` key containing `xrefs: [memory:gone-one,memory:gone-two]` → lint reports gone-one only (the widened gate skips the parsed channel without `refs:`; the body scan can't match after `,`). Coupled: `akm mv memory:old-note memory:new-note` with a citer `xrefs: [memory:other,memory:old-note]` → `rewrote: []`, the ref dangles silently. Spaced forms work; md files and task files with `refs:` are covered by the parsed channel.

**Fix:** add `,` to `REF_BOUNDARY_PREFIX_CLASS_SRC` (it's already a slug terminator, so `a,b` splits cleanly), or run the parsed-YAML xref pass for task files without `refs:` too. See [round-2 doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md) R2-5.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,778 @@
...
+          "cannot be moved on its own without breaking its belief-inheritance coupling to the base memory. " +
+          `Rename the base ref instead (akm mv ${baseRef} <new-name>); the twin moves with it.`,
+        "INVALID_FLAG_VALUE",
+      );
+    }
+
+    // The target may be a bare name ("projectA/new-note") or a ref-shaped
+    // spelling. Parsing the bare form through the same ref grammar gives it
+    // identical name validation (traversal, null bytes, absolute paths).
+    const target = parseAssetRef(targetArg.includes(":") ? targetArg : `${source.type}:${targetArg}`);
```
fwdslsh-dev, line 682: 🟠 **R2-6 (new defect) — trailing-slash target renames the asset to a hidden dotfile and drops it from the index.** `validateName`'s `path.posix.normalize` preserves a trailing slash and the segment check only rejects `.`/`..`, so the empty last segment of `bar/` passes — and mv adds no empty-segment check of its own.

Repro (e2e at HEAD): `akm mv memory:foo bar/` → exit 0, `to:"memory:bar/"`; the file lands at `memories/bar/.md` (dot-prefixed — the index walker skips it); all citers rewritten to `memory:bar/`; `akm show memory:bar/` fails ("unsupported file type or unrecognized layout"); `akm lint` reports **zero** missing-refs (the resolver maps `memory:bar/` to the existing `memories/bar/.md`); after `akm index --full` the asset is gone from search entirely — history orphaned, no warning at any step. Same via backslash (`bar\` normalizes to `bar/`).

**Fix:** reject empty path segments in `validateName` (or in mv's target validation): any segment `=== ""` after normalization → `UsageError`. Add trailing-slash/backslash cases to the mv tests. See [round-2 doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md) R2-6.

---

src/commands/mv-cli.ts
```diff
@@ -0,0 +1,778 @@
...
+  stashDir: string,
+  relPath: string,
+  refType: AkmAssetType,
+  refName: string,
+): string | null {
+  const resolved = resolveRefPathInStash(relPath, refType, refName, stashDir);
+  if (!resolved) return null;
+  if (resolved.endsWith(".derived.md") && !refName.endsWith(".derived")) return null;
+  if (path.basename(resolved) === "SKILL.md" && !relPath.endsWith(`${path.sep}SKILL.md`)) return null;
+  if (path.resolve(resolved) === path.resolve(stashDir, relPath)) return resolved;
```
fwdslsh-dev: 🟠 **R2-7 (new defect) — wrong-case source ref defeats the canonical gate on case-insensitive filesystems (macOS/Windows defaults).** `resolveRefPathInStash` returns `path.join(root, relPath)` verbatim on a direct `existsSync` hit, so this comparison checks the user-cased string **against itself** — it can only reject fallback spellings, never a case alias. On case-insensitive filesystems `existsSync` matches insensitively, so `akm mv memory:Foo bar` (file on disk `memories/foo.md`) proceeds as canonical.

Repro (HEAD; case-insensitive FS semantics emulated via preload, with a case-sensitive control proving `existsSync` is the sole guard): exit 0, `rewrote: []` (patterns have no `/i`; the aliasScan rescue byte-compares against the user-cased `oldPathResolved`, :209), every `memory:foo` citer dangles (post-move lint confirms), `rekeyIndexForMove` derives `…:memory:Foo` which misses the BINARY-collated stored key, the `strandedRow` probe misses the same way — `utilityPreserved:true` reported while the real row is stranded as a ghost and a duplicate row is minted (the exact history reset this gate's own docstring exists to prevent). `rekeyStateDbForMove` strands salience/outcome rows identically.

**Fix:** after resolution, compare against the on-disk casing — e.g. `fs.realpathSync.native(resolved)` must equal the ref-derived path byte-wise; on mismatch fire the existing alias-rejection error naming the canonical ref. See [round-2 doc](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md) R2-7.

---

fwdslsh-dev (commented):
## Round-2 review — fix commits `3ca2779..6a1d844`: 12/15 fixed, 7 fresh verified findings

Re-reviewed at `6a1d844`. Method: every round-1 finding's original repro re-run against the fix commits, plus a fresh blind-spot sweep; each fresh candidate adversarially verified (**7 survived, all CONFIRMED with empirical repros at HEAD; 0 refuted**).

**The fix round is solid.** All five original 🔴 issues (#1 secret corruption, #2 history loss, #3 YAML-workflow corruption, #4 FK rollback, #5 salience re-key) are closed as reported — verified end-to-end, including secret bytes intact, `.yaml` programs untouched, events surviving `--full` reindex, and the embeddings-FK regression test. The grammar constants are now genuinely shared between lint and `mv`, and the below-cut items (HTML-comment capture, surrogate split, workflow write-path indexing) are all properly fixed.

Full round-2 report: 📄 **[code-review/pr-715-stash-conventions-review-round2.md](https://github.com/itlackey/akm/blob/claude/code-review-xhigh-pr-6pz5j7/code-review/pr-715-stash-conventions-review-round2.md)** (includes the per-finding fix-assessment table).

### What remains — 7 fresh findings, each an inline comment with repro + fix

| # | Kind | Where | Issue |
|---|---|---|---|
| R2-1 | 🟠 new defect (in #2's fix) | `db.ts:531` | `mv` adopts a deleted stranger's detached usage events at the target ref — next `--full` relinks them (verified utility 0.30/showCount 5 for zero real usage) |
| R2-2 | 🟠 incomplete #5 | `mv-cli.ts:463` | per-table `catch {}` swallows real state.db errors (e.g. `SQLITE_BUSY`) — salience stranded with exit 0 and no warning |
| R2-3 | 🟡 incomplete #12 | `stash-cli.ts:250` | stdin doc with its own frontmatter + `--xref` still random-slugs; one-line fix via `parseFrontmatter(content).content` |
| R2-4 | 🟠 regression (from #14's fix) | `memory-belief.ts:82` | `writeContradictEdge`'s guard ignores `beliefState` — a scalar-edge file with missing demotion is now a permanent no-op (base repaired it) |
| R2-5 | 🟡 residual #10/#15b | `base-linter.ts:167` | refs after a bare comma (`[memory:a,memory:b]`) invisible to both lint and mv — `,` ends a slug but isn't a boundary |
| R2-6 | 🟠 new defect | `mv-cli.ts:574` | `akm mv memory:foo bar/` → hidden `memories/bar/.md` dotfile; asset silently vanishes from the index, lint sees nothing |
| R2-7 | 🟠 new defect | `mv-cli.ts:299` | wrong-case ref on macOS/Windows defeats the canonical gate — zero citers rewritten, index re-key misses, `utilityPreserved:true` anyway |

**Suggested triage:** R2-1 / R2-2 / R2-4 first (silent-history / silent-failure class), then the input-validation pair R2-6 / R2-7, then the two narrow residuals R2-3 / R2-5.

*Automated review — filed as comments; the blind-spot sweep over the rest of the PR surfaced nothing beyond these.*

---

itlackey:
All 7 Codex findings verified true (each reproduced with the real CLI before fixing) and fixed in `e4ea514`, with 9 pre-fix-red regression tests:

- **C1/C2** — `akm mv` canonicalizes its own source *and* target arguments: `.md`-alias spellings no longer strand utility/salience/outcome history behind a non-canonical `entry_key` or mint a duplicate index row. (One refinement on C1: canonical citers were already rewritten by the resolver pass; the damage was confined to the re-keys, patterns, and report.)
- **C3** — target parent is validated/created *before* citer rewrites; a blocked target now fails with every citer byte-identical.
- **C4** — the state.db re-key swallows only the missing-legacy-table case; real failures surface in the report's `warnings`.
- **C5** — confirmed: `parseAssetRef`'s `TYPE_ALIASES` (`environment:` → `env:`) and `local//` acceptance meant raw spellings persisted that the lint scanner would never match. `--xref`/`--supersedes` now persist canonical `type:name` and dedupe across alias spellings.
- **C6** — the resolver-backed rewrite retries unresolved tokens with trailing punctuation stripped (`See memory:old.` now rewrites, punctuation preserved; genuinely dotted names that resolve win first). Lint's shared grammar is untouched.
- **C7** — workflow YAML programs joined the citer scan (they carry refs in instruction text); *moving* workflow refs stays rejected.

Full unit suite 26,288 pass; mv/xref/supersedes/lint/contract suites green; docs/cli.md updated.

---

itlackey:
Round-2 findings are resolved in `314cbf8` (note: your round-2 review raced my Codex-findings commit `e4ea514`, which explains the one already-fixed item):

| # | Status | Resolution |
|---|---|---|
| R2-1 | fixed | re-key now evicts detached orphan `usage_events` at the target ref (bare + origin-qualified spellings) before the old→new rewrite — live-asset-wins, matching the state.db collision policy |
| R2-2 | already fixed by `e4ea514` (Codex C4, same defect) | re-verified with your repro shape: a held `BEGIN IMMEDIATE` now yields `warnings: ["state.db salience re-key failed (asset_salience: database is locked)…"]`, exit 0 |
| R2-3 | fixed | slug inferred from `parseFrontmatter(content).content`; the no-flag path derives the same slug, and the frontmattered-stdin case is pinned in the integration test |
| R2-4 | fixed | `writeContradictEdge` guard is beliefState-aware again (mirrors the sibling; `archived` never weakened) — thanks for catching a regression our own #14 fix introduced |
| R2-5 | fixed | `,` added to `REF_BOUNDARY_PREFIX_CLASS_SRC` — no-space flow lists now flag in lint and rewrite in mv |
| R2-6 | fixed | mv rejects empty path segments (trailing `/` and `\`) before any write |
| R2-7 | fixed | on-disk casing verified via per-segment directory-listing comparison — deliberately not `realpathSync.native`, which resolves symlinks and would false-mismatch symlinked stash roots (e.g. macOS `/tmp`); wrong-case refs get the alias rejection naming the canonical ref |

Each live finding was reproduced at HEAD before fixing, with pre-fix-red regression tests (+250 test lines). Full gate green over the combined stack: `bun run check` exits 0 — 26k+ unit / 766 integration pass, contracts and lint guards clean.

---

PR #719, opened 2026-07-13
akm 0.9.0 — architecture refactor (Wave 1 complete; Wave 2 in progress)

itlackey:
## Summary

This PR carries the akm 0.9.0 architecture refactor, executed chunk-by-chunk against the plan in `docs/design/akm-0.9.0-*` and the manifest `docs/design/akm-0.9.0-chunk-manifest.json`. Every work item is a self-contained commit with its own gate-green verification; per-chunk execution ledgers live under `docs/design/execution/chunk-*/`, and the honest whole-PR actuals (vs the plan's projections) are published in `docs/design/execution/actuals-0.9.0.md`.

> This description is maintained live as work lands. It supersedes the original "docs-only plan" description — the plan documents are still here, but the branch now contains the implementation.

## Status — all planned chunks, close-out groups, and post-sweep rulings COMPLETE ✅

- **Wave 1** — chunks 6, 7 residuals, 9 (import cycles 107 → 0 baseline by close, typed errors, RunContext unification, god-module decomposition).
- **Wave 2** — chunks 0b, 1, 1.5, 2, 3, 4, 5, 6.5, 8, 10: adapter model + util home, taxonomy-global deletion, wiki asset-type retirement, IndexDocument + 0.9.0 ref grammar, activation policy, the three-DB merge with journaled cutover + config/lockfile split, contract-surface sweep.
- **Close-out (user-ruled groups)** — Group A (linter consolidation −454, config-schema decomposed into 12 modules byte-identically, scored/enumerate filter unification, ambient-config sweep); the 8 spec'd format-family adapters built to their specification goldens (237/0 conformance); Group C (F3 dispositions, derived_from grammar flip, semantic-e2e re-home, `--auto-accept` upgrade map, and `noUncheckedIndexedAccess` enabled tree-wide — 1,902 violations → 0).
- **Post-sweep rulings (2026-07-21)** — the adapter registry wired into production as the spec §12.6 static frozen `BUILTIN_ADAPTERS` list with a genuine ordered `looksLikeRoot` probe (`akm` no-match fallback kept as a recorded deviation from §1.2(3)); all 13 pre-0.9.0 ref-grammar residue sites purged per D-R3 (help/error/LLM-schema text, task-ref parsing, loud proposal filters, reader-traced durable-sentinel flips); the dead `akm wiki` output surface deleted (−176 LOC, zero-importer proven) and every shipped hint/doc/skeleton seed rewritten to the llm-wiki bundle story.

Full gate at HEAD: lint 0 · tsc 0 (with `noUncheckedIndexedAccess`) · unit 2806/0 (216/216 files) · integration 4585/0 (351/351 files) · CI green (CI, akm-eval smoke, Workflow Syntax).

## Remaining before merge

- rc.6 version bump at cut time (deferred to the actual cut by decision).
- Owner ruling on residue item 2 (retired shapes: `resolveStashDir` silent `stashDir` reader, proposal legacy-importer relocation, `entry_key` reader, gate-counter plumbing) and the migrate-apply restore-path WAL-conversion hardening — 0.9.0-vs-0.9.1 disposition.
- Recorded 0.9.1 deferrals: three-verb rename, item_links, L0/L1/L2, jsonColumn, Group B remainder, Group D, adapter scan-engine swap + §9.4 selection persistence, `okf` probe fallback.

## Flagged for review (decisions made autonomously, documented in the ledgers)
1. **Net-LOC actuals vs projections** — published plainly in `actuals-0.9.0.md` (src net +7,042 vs projected −11k..−13.7k, with drivers).
2. **Improve-scoped deferrals (WI-9.10d/e)** — ambient-DI seams retained per the binding di-seams plan.
3. **`task` recognition defect fixed** (`tasks/*.yml` unrecognized since 0.8.0) + `applyTaskMetadata` YAML fix — investigated-as-oversight, approved, folded in.
4. **Census correction** (chunk-0b anchors D.1) — scored/enumerate derived-twin asymmetry is candidate-pool construction; empirically verified.

## Test plan
Each work item lands its named gates green (tsc, biome, import-cycle ratchet at an EMPTY baseline, fn-size ratchets, frozen-golden sha256 presence lint, targeted + affected suites). Full sharded unit + integration batteries run at every landing.

https://claude.ai/code/session_019fAVPbUo6mA43wvnx5N1iY


fwdslsh-dev:
Ten explore agents completed independent source, test, documentation, tooling, static-analysis, and coverage sweeps. src/migrate/** was excluded.
High Severity
1. Retired ref grammar remains active across runtime paths.
- Task prompts recognize agent:name but not canonical agents/name: src/tasks/parser.ts:344-365, src/tasks/validator.ts:73-77.
- Proactive-maintenance scoring expects type:name, causing canonical refs to lose type weights: src/commands/improve/proactive-maintenance.ts:100-158.
- Feedback reconstructs rejected lesson refs: src/commands/feedback-cli.ts:381-400.
- Sessions and tasks emit session:* and task:*: src/commands/improve/session-asset.ts:270-308, src/commands/tasks/tasks.ts:243-383.
- Consolidation prompts request legacy refs that validation discards: src/assets/prompts/consolidate-system.md:6-16.
- Runtime dual readers remain in workflows, usage events, mv, and standards resolution.
2. Env and secret mutations bypass canonical write-target handling.
- findEnvSource() selects the first read source and ignores --target, defaultWriteTarget, writability, indexing, and Git commit boundaries: src/core/env-secret-ref.ts:46-65.
- Mutation paths call filesystem helpers directly: src/commands/env/env-cli.ts:353-492, src/commands/env/secret-cli.ts:248-269.
- Documented env:prod and secret:key become literal colon-bearing names instead of canonical refs.
3. Ref resolution has competing implementations.
- Canonical resolveRef() has no production caller: src/core/asset/resolve-ref.ts:109-136.
- Show, feedback, graph, and index lookup resolve refs independently.
- Short database refs use unordered suffix matching with LIMIT 1: src/storage/repositories/index-entries-repository.ts:676-731.
- parseQualifiedRefInput() also permits source locators such as npm refs where bundle slugs are expected.
4. Configured adapters are recorded but ignored.
- Config accepts and installation records adapter IDs, but indexing always uses akmAdapter: src/indexer/indexer.ts:929-935.
- The intended adapter path is isolated in the production-unreachable src/core/adapter/scan-component.ts.
- Non-AKM bundles can therefore be skipped, misclassified, or assigned false provenance.
5. Release validation is both bypassed and partly broken.
- .github/workflows/release.yml:36-88 publishes without bun run check or release:check.
- tests/release-check.sh:96,105-106 references three moved/nonexistent test paths.
- .github/workflows/release-workflow-syntax.yml watches the wrong workflow-contract path.
6. Setup mutates external state before confirmation.
- Scheduled tasks and OS scheduler entries are created before config confirmation/save: src/setup/setup.ts:464-594.
- Task setup reloads persisted config instead of using the draft: src/commands/tasks/tasks.ts:783-785.
- Cancelling setup can leave task files and scheduler registrations behind.
7. Managed source reads, refreshes, and writes disagree on the authoritative location.
- Reads prefer lockfile localRoot, indexing refreshes re-derived providers, and writes derive another cache path.
- Relevant paths: src/indexer/search/search-source.ts:127-138,326-359, src/core/write-source.ts:559-576.
- Source update/add/clone operations also delete or persist old state before replacement succeeds.
8. Network boundaries have material safety gaps.
- Website SSRF checks do not resolve DNS before allowing hostnames: src/sources/snapshot-fetchers/website-ingest.ts:644-732.
- fetchWithTimeout() clears its timeout after response headers and overwrites caller cancellation: src/core/common.ts:297-358.
- Archive downloads have no body-duration or byte cap.
9. LLM/index status can claim success after failure.
- Enrichment can be gated off internally yet still be marked enriched and cached: src/indexer/indexer.ts:1758-1782.
- sqlite-vec insertion failures are swallowed while readiness is inferred from BLOB counts: src/storage/repositories/index-vec-repository.ts:113-146.
- Search mode telemetry infers semantic use from elapsed milliseconds rather than an explicit result.
10. Setup reintroduces unsafe LLM defaults.
- Wizard-created connections set maxTokens: 1024: src/setup/steps/connection-shared.ts:22-35.
- Non-local indexing defaults concurrency to 4.
- Both contradict the repository’s intended defaults of no token cap and concurrency 1.
Confirmed Removal Candidates
- src/indexer/manifest.ts: production-unreachable, maintained only by tests/integration/manifest.test.ts.
- src/core/adapter/scan-component.ts: explicitly unwired but 100% covered by test-only use.
- src/integrations/harnesses/opencode-sdk/index.ts: orphan barrel.
- src/core/config/config-io.ts:262-373: unused unified-diff/LCS subsystem.
- src/indexer/search/search-hit-enrichers.ts:110-140: extension registry that can never contain an extension.
- src/workflows/program/parser.ts:933-981: stale dynamic loader for an existing sibling module.
- src/commands/url-checker.ts: production always supplies empty bodies, so it cannot find URLs.
- src/core/ripgrep/**: initialization installs an unused, unverified binary through curl.
- Five src/assets/help/*.md, staleness-detect-system.md, and graph-refresh-weekly.yml: shipped without production consumers.
- scripts/migrations/*: described as obsolete, still copied into distributions, and excluded from normal lint/typecheck.
- Static analysis found 24 zero-reference runtime exports and 5 zero-reference exported types.
Superseded Implementations
- The retired metadata-contributor pipeline remains test-maintained beside the live adapter fold: src/indexer/passes/metadata.ts, src/indexer/passes/metadata-contributors.ts.
- Task subprocess handling is weaker than the existing agent process-group implementation: src/tasks/runner.ts:235-315 versus src/integrations/agent/spawn.ts:501-664.
- Setup runs bun add @huggingface/transformers against the installed package despite it already being optional: src/setup/semantic-assets.ts:49-81.
- Ripgrep installation shells out to curl despite an existing in-process downloader.
- Two Linux-specific shell test runners duplicate about 100 lines of sharding machinery each.
Test Hygiene
- Approximately 15 tests exercise removed provider-result merging; mergeSearchHits is production-dead.
- The legacy-ref ratchet excludes 70 files containing 653 raw legacy-shaped tokens.
- manifest and scan-component have high coverage solely because tests preserve unreachable production modules.
- The semantic E2E documentation points at the wrong path and its key deploy assertion is conditional.
- Structural contract tests inspect documentation that CI explicitly ignores on docs-only changes.
- Strong tests for directory staleness, YouTube fetching, lesson lint, and feedback valence were previously deleted while production code remained.
- Vault terminology and expired compatibility aliases are still primarily sustained by tests.
Coverage
- 3,888 test executions across broad and focused coverage profiles passed with zero failures.
- Production-only broad coverage outside src/migrate/**: 68.19% functions, 58.82% executable lines.
- Completely uncovered runtime areas:
- src/sources/snapshot-fetchers/youtube.ts: 0/13 functions.
- src/sources/snapshot-fetchers/registry.ts: 0/2 functions.
- src/sources/providers/sync-from-ref.ts: 0/2 functions.
- src/registry/semver.ts: 0/7 functions.
- Important weak areas include task validation, website redirect/security branches, setup failure paths, and directory-staleness edge cases.
Complexity
- 541 TypeScript modules, 142,180 source LOC, 5,848 functions, and zero import cycles.
- 13 functions exceed the repository’s 220-line threshold; 59 have complexity proxy scores above 30.
- Largest hotspots include src/commands/improve/consolidate.ts, preparation.ts, src/indexer/indexer.ts, and src/commands/proposal/repository.ts.
- Highest function complexity includes validateStashEntry, formatWorkflowBriefPlain, foldRecognizedMetadata, and workflow execution orchestration.
Documentation Drift
- docs/cli.md documents removed commands, flags, shapes, and auto-accept behavior.
- Current help, prompts, skeleton assets, and agent guides teach rejected type:name refs.
- AGENTS.md and architecture docs still describe removed SourceProvider.init; the actual interface is name, kind, path, and optional sync.
- Storage docs and a contract test still pin retired workflow.db.
- Agent hint content differs between source checkouts and npm installations.
- Privacy wording says no data reaches third parties while configured LLM, registry, source, and GitHub requests do so.

Agents used the dedicated grep engine, TypeScript compiler API, AST reachability/reference analysis, complexity scoring, runtime parser probes, Git history, and Bun coverage. No files were changed; the worktree remains clean.

itlackey:
## Triage of the ten-agent review (verified against `804e456a`)

Every high-severity finding was independently re-verified against the current head before acting. Status per finding, with reasoning for anything invalidated or deferred.

### Being addressed now

- **#1 Retired ref grammar in runtime paths — confirmed, in progress.** Verified the worst of it: `resolvePromptSource` recognizes only colon refs as assets, so a canonical `commands/release` task prompt silently becomes *inline text*, while a colon ref that is recognized then fails `parseRefInput` in the validator — broken in both directions. `refType` in proactive-maintenance returns `""` for canonical refs and drops type weights. `session:*`/`task:*` durable emissions and the consolidate prompt confirmed. Fixes are in flight with reader-traced lockstep flips. **One carve-out:** the "runtime dual readers remain in workflows, usage events, mv, and standards resolution" bullet is *deliberate* — those readers carry explicit D-R3 migration-window comments (pre-cutover durable rows can still reach them) and are scheduled for removal with the migrator at 0.10.0. They are not being changed.
- **#5 Release validation — fixed in `ca4e6313`.** Confirmed: `release.yml` had no check step; `release-check.sh` referenced `tests/workflow-release.test.ts` and `tests/install-script.test.ts` (moved to `tests/integration/` in the integration split) and `tests/setup.test.ts` (replaced by the `tests/setup/` suite); the Workflow Syntax trigger watched the moved path. `release.yml` now runs `bun run check` on the tagged commit before build/publish, and all paths are repaired.
- **#6 Setup mutates external state before confirmation — confirmed, in progress.** The `scheduled-tasks` step (last in the wizard) creates task files and OS scheduler entries before final persist, and the task-setup path reloads persisted config instead of the draft. Fix reorders side effects after confirm/persist and threads the draft.
- **#8 Network boundaries — confirmed, in progress.** `fetchWithTimeout` clears its timer once headers arrive (unbounded body window); archive downloads lack byte/duration caps; the SSRF checks validate hostnames without resolving them. Fixes in flight. Note: Bun's `fetch` offers no custom-lookup connection pinning, so DNS-rebinding TOCTOU can only be *narrowed* (resolve-and-validate every address + re-validate on each redirect hop); the residual window will be documented in-code.
- **#9 Status truthfulness — confirmed, in progress.** `quality = "enriched"` is set and cached unconditionally after the enrichment call; vec-insert failures are swallowed with readiness inferred from BLOB counts; search-mode telemetry infers semantic use from elapsed time. All three being fixed (honest degradation, not hard failure — semantic search remains optional).
- **#10 Setup LLM defaults — confirmed, in progress.** `maxTokens: 1024` is hardcoded where the schema deliberately leaves the field optional. The agent is establishing the authoritative intended defaults (schema/runtime/git history) before changing wizard behavior, same for the indexing-concurrency default.
- **Removal candidates, superseded implementations, documentation drift — verify-then-act in progress.** Each candidate requires fresh zero-importer proof before deletion; `mergeSearchHits` + its dead tests included. Docs pass covers `docs/cli.md` accuracy, `AGENTS.md` `SourceProvider` shape, the workflow.db storage doc + contract test, remaining `type:name` teaching, and the privacy wording (which will state precisely that akm sends nothing to its authors but user-configured endpoints necessarily receive the requests the user configures).

### Invalidated or deferred, with reasoning

- **#4 "Configured adapters are recorded but ignored" — already adjudicated, no new action.** Partially stale: as of this branch the registry is the spec §12.6 static frozen `BUILTIN_ADAPTERS` list and `detectAdapterId` genuinely runs the ordered `looksLikeRoot` probe in production. What remains — per-file recognition still flowing through `akmAdapter` rather than the per-component adapter (`indexer.ts:929-935`), with `scan-component.ts` as the future consumption site — is a **recorded deviation**: it is documented in `scan-component.ts`'s PRODUCTION STATUS comment and listed in this PR's "Recorded 0.9.1 deferrals" (scan-engine swap + §9.4 selection persistence). The finding restates a known, deliberately-sequenced boundary. `scan-component.ts` is likewise excluded from the removal sweep for the same reason.
- **#2 Env/secret write-target handling — confirmed real, deferred to 0.9.1 by prior ruling.** `findEnvSource` does ignore `--target`/`defaultWriteTarget`/commit boundaries (verified). But the 0.9.0 planning refinements explicitly ruled *"env/secret handling stays unchanged"* for this release — the write-target/Git-boundary integration is a behavior change to a security-sensitive surface and belongs in 0.9.1 with its own tests. The one piece landing now is doc accuracy: docs teaching `env:prod`/`secret:key` spellings are corrected in the docs pass.
- **#3 Competing ref-resolution implementations — deferred to 0.9.1.** Verified: canonical `resolveRef()` has no production caller and commands resolve via `parseRefInput` + per-command lookup; the LIMIT-1 unordered suffix match in `index-entries-repository` is real. Consolidating every command onto one resolver re-plumbs the lookup path of the whole CLI — exactly the kind of wide, behavior-adjacent change that should not land days before an rc cut. Tracked as the resolver-consolidation item for 0.9.1 (the suffix-match determinism fix rides along with it, since its blast radius is the same lookup path).
- **#7 Managed-source authority disagreement — confirmed, deferred to 0.9.1 as a design item.** The read side is the *deliberate* §10.2 lock-first design (`resolveEntryContentDir`: lock `localRoot` first, provider fallback for migrated entries). The genuine defect is that the write path (`pathFromConfiguredSource`) re-derives the git cache root instead of sharing that resolver, and refresh has a third view. The right fix is one authoritative content-root helper shared by read/refresh/write — a design-reviewed unification, not a patch; scheduled for 0.9.1.
- **Test hygiene / coverage / complexity — 0.9.1 backlog**, with two exceptions handled now: the `mergeSearchHits` dead code + its ~15 tests (removal sweep), and the release-check path rot (fixed above). The legacy-ref ratchet's 70-file exclusion list, the coverage gaps (youtube/registry/sync-from-ref/semver at 0%), and the 13-function fn-size hotspots are real and belong to the 0.9.1 hygiene wave; none are regressions introduced by this PR.

Fix commits will land on this branch behind the full check battery as the in-flight work completes.

itlackey:
## Review-response landing (`ca4e6313..1a954ce9`, 17 commits)

All valid findings from the ten-agent review are now addressed on this branch. Full local gate: lint 0 · tsc 0 · unit 2797/0 (215/215 files) · integration 4584 pass with a single failure that is **environmental to the build machine** (fails identically at the pre-stack, CI-green base — a `git:` source probe hangs on this host); CI on this push is the arbiter for it.

### What landed
- **Release pipeline (finding 5)** — `release.yml` now runs `bun run check` on the tagged commit before publish; the three rotted `release-check.sh` paths and the Workflow Syntax trigger repaired.
- **Runtime ref grammar (finding 1)** — task prompts now recognize canonical `[bundle//]subdir/name` asset refs and reject colon grammar with a typed hint (previously canonical refs silently became inline text); `refType` derives from conceptId so proactive maintenance keeps type weights; `session:*`/`task:*` emissions flipped with reader traces; the consolidate prompt was confirmed **silently broken** (its `memory:<name>` op refs were exact-match-dropped as hallucinated) and is now canonical; plus a found-along-the-way bug — `--applied-to` lesson credit never matched because the lookup was built in the old grammar.
- **Setup (findings 6 + 10)** — task files and OS scheduler entries are created only after the final confirm/persist (cancelling leaves nothing behind; regression-tested), the task step now sees the confirmed config, and the wizard's hardcoded `maxTokens: 1024` is gone (authority: the "remove artificial limits" change and the client's own no-default contract).
- **Network boundaries (finding 8)** — body reads are now deadline-bounded (the fetch timer only ever covered headers), archive downloads have byte + duration caps, website SSRF checks resolve DNS and validate every address (fail-closed, re-checked per redirect hop; the un-pinnable-socket residual is documented in-code), and a real bug fell out: the remote embedder's cancellation signal was being silently dropped.
- **Status truthfulness (finding 9)** — enrichment is marked/cached only on genuine success via a tri-state outcome (no more permanent-skip poisoning after failures), vec insert failures are counted and surfaced with readiness reflecting actual outcomes (honest degradation, search still falls back), and search-mode telemetry carries an explicit `usedSemantic` flag instead of guessing from elapsed time.
- **Removals** — **−2,403 LOC** of verified-dead code: indexer/manifest, the opencode-sdk barrel, config-io's diff/LCS subsystem, the never-populated search-hit-enricher registry, the stale expression loader, the entire unused ripgrep-install subsystem (curl, no checksum, zero callers), orphaned shipped assets, obsolete migration scripts, and production-dead `mergeSearchHits` + its tests.
- **Docs drift** — cli.md realigned (deprecated `--auto-accept`, phantom `--min-retrieval-count` removed, remaining old-grammar refs fixed), `SourceProvider` docs match the real interface, storage docs + the contract test re-pinned off retired `workflow.db`, and the privacy doc now states precisely that akm has no telemetry of its own while user-configured endpoints necessarily receive the requests the user directs.
- **Bonus fix** — the improve dead-URL checker was inert (its only caller passed `body: ""`); it now reads the document body from the pre-resolved file path.

### Corrections to review claims (found during verification)
- **Indexing concurrency default**: the review said the repo intends 1; repo authority says otherwise — `getDefaultLlmConcurrency` deliberately defaults to 4 for non-local endpoints with justifying comments post-dating the concurrency-1 change. Left as-is.
- **Hints divergence between checkout and npm install**: false — the observed divergence was a stale local gitignored `dist/`; `prepublishOnly` rebuilds and ships `src/assets/hints` byte-identical.
- **Metadata-contributor pipeline "retired"**: inverted — `metadata-contributors.ts` is the *live* recognition path (13 active registration sites); the unreferenced artifact is the adapter-side port. Nothing deleted.
- **`graph-refresh-weekly.yml` "no consumers"**: false — it is statically imported as the bundled graph-refresh task; kept.
- **`fetchWithTimeout` "overwrites caller cancellation"**: not quite — abort reasons cannot overwrite each other; the real defects were the signal's scope ending at headers and one caller passing `signal` where it was silently dropped. Both fixed.
- **`url-checker` "cannot find URLs, remove it"**: it is a genuine feature whose caller lost its body plumbing — fixed rather than deleted.
- **`semantic-assets` install step "redundant"**: it is the guarded auto-install fallback for a genuinely absent optional dependency; kept.
- **`scan-component.ts` removal**: excluded by prior ruling — it is the documented 0.9.1 scan-engine consumption site.

Deferred items (env/secret write-target integration, resolver consolidation, managed-source authority unification) remain as stated in the triage comment above.

itlackey:
## Subprocess-management consolidation (`846fed4cb87519f81db61031fbec860fdb46eb30`)

Follow-up to the review's "Superseded Implementations" item on task subprocess handling — which the landing comment above omitted (a triage miss, corrected here). Investigation confirmed the claim and found it understated: the task runner's command path had no process group (timeout orphaned grandchild processes), **no SIGKILL escalation** (a SIGTERM-ignoring child wedged the run forever — the stream reads and `exited` never resolve), unbounded output slurping, and no test seam.

**Shape landed:** a managed-subprocess primitive (`src/core/subprocess.ts`, a runtime-boundary-respecting leaf) now owns process-group spawn, the SIGTERM→SIGKILL ladder, bounded capture, timeout/abort, and injectable spawn/timer seams — extracted from `spawn.ts`, not written new. Three consumers: `runAgent` (contract byte-identical, shrank 738→474 lines, pinned by 1052+ assertions across the agent/harness/native-executor suites), the task-command runner (weak inline copy deleted; log shapes, 143-on-timeout, and env stamping preserved exactly; gains group-kill + escalation + a deterministic escalation test), and the two previously-unbounded setup probes (`ollama list` at 10s, `bun add @huggingface/transformers` at 10min). Task commands deliberately do **not** inherit agent semantics — `runAgent` remains the sole harness-facing agent surface, and workflows reach subprocesses only through it.

Gate: lint 0 · tsc 0 · unit 2804/0 (216/216) · integration 4585 pass with only the previously-documented build-host-environmental failure; CI on this push arbitrates.

itlackey:
## Follow-ups on the review-claim corrections (`c0f6af07`)

**LLM enrichment concurrency (owner ruling):** now defaults to **2 for remote endpoints, 1 for local model servers**, `llm.concurrency` overriding both — code, comments, `docs/technical/indexing.md`, and a new pinning test. Writing the test surfaced a latent bug: `URL.hostname` keeps IPv6 brackets, so `http://[::1]` endpoints were misclassified as remote; loopback detection now strips them.

**Record correction — metadata-contributor pipeline:** the landing comment above said the reviewer had this inverted; fact-tracing shows the **reviewer was right and our rebuttal was wrong**. The only call into `applyMetadataContributors` sits inside `buildEntryFromFile`, whose sole caller `generateMetadata` has zero production importers — the live scan path is the adapter fold (`drain-dir → akmAdapter.recognize → akm-metadata.ts`), which reuses only the shared field helpers. The async contributor registry is production-dead code sustained by the fold's parity tests. Disposition (retire now and re-anchor the parity tests onto the fold, vs. 0.9.1) is with the owner.

Verified from code in the same pass: the scan-component deferral reasoning holds (the indexer is a per-directory incremental engine — dir-staleness, per-dir drain/persist/LLM-cache — while `scanComponent` is a per-component stream; the swap is a scan-architecture rewrite plus the not-yet-built sub-mount/manifest pieces, not a call-site swap), and the three deferred items re-sized honestly: env/secret write-target routing is a real user-hittable correctness gap (early 0.9.1, per the v0.3 spec's no-new-env/secret-machinery staging), resolver consolidation touches ≥8 independent command lookup paths (deferral stands; the `LIMIT 1` determinism fix could be pulled forward alone), and the managed-source authority unification is smaller than first triaged (`pathFromConfiguredSource` has no callers outside write-source.ts) — a focused early-0.9.1 item.

Gate at `c0f6af07`: lint 0 · tsc 0 · unit 2809/0 (217/217) · integration 4585 pass + the previously-documented build-host-environmental failure; CI arbitrates.

itlackey:
## Owner-ruled refactor completion (`c0f6af07..e9327636`, 7 commits)

The items previously labeled 0.9.1 deferrals were ruled 0.9.0 work that fell through the cracks, and are now complete:

- **Adapter dispatch is live** — the indexer resolves each component's detected adapter (`adapterForId`) instead of hardcoding `akm`; unknown ids skip the component with a warning (spec §4). Pinned by a production-path test: the llm-wiki fixture indexes under llm-wiki's own types, never akm's.
- **Adapter-owned filtering (owner ruling)** — `shouldIndexStashFile`'s akm-stash policy (sensitive-marker env/secret skips, legacy `vaults/`, wiki infra) moved into the akm adapter as path-based recognize-abstention; the drain no longer pre-filters, so every adapter sees its bundle's walked files and decides itself. Bytes-never-read for sensitive files proven by a context whose `content()` throws.
- **Manifest removed from the spec entirely; sub-mounts replaced by adapter-owned processing** — recorded as the v0.4 amendment (2026-07-21 owner ruling) in the format-neutral spec (§9.2 one bundle = one component = one adapter; §9.3 nested-root subtraction removed; §14.2 rewritten to the real engine) and the 0.9.0 adapter spec (§1.2 steps 1/4). `scanComponent` — the unwired second engine — deleted, with conformance retargeted onto walk×recognize directly.
- **Contributor-registry retirement** — the production-dead pipeline (`generateMetadata`/`buildEntryFromFile`/`applyMetadataContributors` + registrations) deleted, −443 LOC; parity tests re-anchored onto the live fold with coverage strengthened (presence AND absence now pinned per context).
- **Short-ref determinism** — the `LIMIT 1` suffix match now orders by installation precedence (`ORDER BY id ASC`, primary stash first). **Record correction #4:** the "≥8 competing resolver implementations" premise — the reviewer's and ours — was wrong: the DB family already shares one resolver and the two filesystem resolvers are legitimately distinct; determinism was the only real defect. `resolveRef()` deliberately remains a pure parse leaf (adopting it would add a layer to produce the same answer).
- **Managed-source authority unified** — one lock-first `lockContentRootFor` shared by reads and writes; a git source with a lock `localRoot` now reads AND writes through the same root (pinned); refresh documented as legitimately provider-deriving (it decides where content gets pulled *into*).
- **Env/secret canonical write-targets** — all six mutations honor `--target` › `defaultWriteTarget` › working stash with writability gating and single boundary commits; every security invariant (0600, values never in output/argv, gitignored `env/` no-op) test-pinned. Reads unchanged.
- **Gremlin caught and fixed before landing** — with dispatch live, the persist layer still re-derived `item_ref` via akm's `stashDirFor` scheme, which would have collapsed non-akm identities (`pages/foo` → bare name). The drain now carries the owning adapter's `conceptId` and persist prefers it verbatim (D-R3); pinned by an `item_ref` fidelity test.

Net for the round: **−566 LOC** in production code. Gate at `e9327636`: lint 0 · tsc 0 · unit 2807/0 (216/216) · integration 4597 pass + only the previously-documented build-host-environmental failure. CI arbitrates. Reindex note: previously-probed non-akm installs reclassify under their own adapter on the next `akm index` (regenerable cache; CHANGELOG updated).

itlackey:
## Final pre-merge fixes (`41da7281`, `c6cab094ea25d2495982509ada12a81217d4e51d`)

**Issue #720 root-caused and fixed** (`41da7281`, CI already green): the migrate-apply WAL→DELETE conversion never failed from contention — it failed because **bun:sqlite's `close()` with unfinalized `prepare()` statements is a zombie-close** (the connection and its WAL shared memory survive until GC; `query()` statements are immune). Proven by a five-probe empirical bisection. Fixed with a scoped `openDatabaseFinalizing` boundary wrapper adopted by the migrate-apply helper connections, cutover probes, and the migration test fixture; the conversion is now **authoritative** — journal-mode read-back, one GC-assisted retry (heals zombies from plain closes anywhere in-process), then an early fail-closed error naming other-process holders, while restore-from-backup is still trivially possible. A global finalize-on-close was tried and honestly reverted: statement caches and worker flows legitimately hold statements across sibling closes. Same commit: the reclassification note promoted to the CHANGELOG's Breaking-changes top, and the two `--force` destroy-before-replace orderings fixed (source update no longer cleans the old root before sync succeeds; clone --force stages-then-swaps).

**Query-time git clone eliminated** (`c6cab094ea25d2495982509ada12a81217d4e51d`): `akm show`'s background auto-index called `ensureSourceCaches`, which unconditionally `sync()`ed every cache-backed source — a §14.3/D11 violation that could hang reads indefinitely on a slow git host (observed live: `git clone` + `ssh git-upload-pack` spawned mid-`show`, wedging 120s on hosts where ssh stalls). The read path now passes `materialize: false`: present caches serve last-known-good with no TTL pull; absent caches skip with one warning naming `akm index`/`akm source update`; materialization remains exclusively in `akm index`, source add/update/sync, and improve's blocking preflight. The previously-hanging test now completes in ~150ms, and the local battery is fully green for the first time on this host: **lint 0 · tsc 0 · unit 2810/0 (217/217) · integration 4599/0 (352/352)** — zero carve-outs.

With this, every item from the ten-agent review, the owner rulings, and the gremlin sweeps is closed. Merge when ready — the release workflow handles the version bump.

Issue #672, opened 2026-06-26
Architecture follow-up: extract salience/outcome repositories and widen the X4 guard


itlackey:
Two follow-ups deferred out of the architecture-refactor D-series (completed in #669/#670/#671). Both are genuine improvements that were intentionally **not** done as part of the decomposition because they require a redesign rather than a behavior-preserving move.

## 1. Decompose the `runImprovePreparationStage` accumulator blocks
After D1c, `src/commands/improve/preparation.ts` holds the prep pipeline, but ~3 internal phases could not be extracted as pure moves:
- **WS-2 Outcome loop**, the **proactive / high-salience selectors**, and the **replay-selection** layer.

They are entangled through `salienceMap` — a **cross-phase mutable accumulator** that the salience-ranking phase, the Outcome loop, and replay-seeding all read and mutate in place. Carving any one out as a function changes behavior unless the accumulator is threaded explicitly.

**Proposed work:** redesign the prep stage to pass the salience accumulator (and the related `pendingForgettingRefs`) explicitly between phase functions (or build it in one pass and pass it read-only downstream), then extract the Outcome loop / selectors / replay as named functions. This is a behavior-changing refactor → needs its own RED characterization tests for the salience ranking + forgetting-candidate outputs first.

## 2. Move command-layer `asset_salience` / `asset_outcome` SQL into repositories, then widen X4
X4 (#671) shipped a **narrow** repository-boundary lint (`scripts/lint-repository-sql.ts`) guarding `src/registry/**` + `src/workflows/runtime/**`. It was deliberately **not** widened to the improve command layer because raw `asset_salience` / `asset_outcome` SQL still lives directly in:
- `src/commands/improve/salience.ts`
- `src/commands/improve/outcome-loop.ts`
- `src/commands/improve/homeostatic.ts`
- (read-side also in `src/commands/health.ts`)

**Proposed work:** extract `src/core/state/salience-repo.ts` + `src/core/state/outcome-repo.ts` (mirroring the D2 repo pattern), move the raw SQL behind them, then add `src/commands/improve/**` (or the specific files) to the X4 guard's `GUARDED_PREFIXES` so the boundary covers the improve layer too. The X4 meta-test already proves the guard fires, so widening is a one-line allowlist change once the SQL is relocated.

## Notes
- These are independent of each other and of the merged D-series; neither is a regression.
[...]

PR #736, opened 2026-07-30
fix(search): remove the R2 salience boost from default ranking (#692)

### LINKED ISSUE #692: search: R2 salience ranking boost needs a config gate and an outcome gate (or removal)
## Problem

The R2 boost (`asset_salience.rank_score` → search/curate ranking, ×[1.0–1.2]) shipped default-on in 0.9.0-beta.53 with **no config opt-out** — only the internal `salienceRankScores: null` API param can disable it. On live data it is both weak and biased:

- `asset_salience` has **zero pack refs** (`%//%` = 0 rows), while externally-installed packs dominate actual consumption (top shows AND curates). The boost can only ever favor self-generated personal assets — worst case ~+7% for an unread knowledge note over an equally-relevant pack asset.
- For 5,560/5,601 non-content-scored rows the encoding term is a per-type constant, so the boost mostly reduces to "is a personal skill/agent".
- Its dominant input (retrieval, w_r=0.60) counts search **impressions** as retrievals with no source filter — a second impression→rank loop on top of the utility EMA.
- The utility EMA contributor (weight 0.5) already delivers usage reinforcement from the same `usage_events`, with full pack coverage — R2 double-counts it for personal assets only.
- Live effect is ~noise: max rank_score 0.356 → max multiplier ×1.071, mean ×1.016.

## Fix options (in preference order)

1. **Remove** the salience contributor from user-facing search ranking (keep `rank_score` for improve-internal selection) until a measured curate-golden-bench delta justifies it.
2. If kept: add a config gate (default **off**), and gate the boost on retrieval-backed outcome signal (`asset_outcome.outcome_score > 0` — column exists, ~541 assets qualify) so it stops double-counting impressions and favoring the unread corpus.

Either way, a config opt-out should exist for a term that touches every query on the hottest read path.

https://claude.ai/code/session_018YjU1mGiShdDP4qFW7p4mv

itlackey:
## Summary

Fixes #692.

- **Removes** the R2 salience ranking contributor from *default* user-facing ranking. `salienceRankingContributor` is dropped from `defaultUtilityRankingContributors` (`src/indexer/search/ranking-contributors.ts`), which is now the one-entry list `[utilityRankingContributor]`. No config gate was added — a key for a term being removed would be dead surface the upcoming 1.0 contract freeze would have to carry.
- **Deletes** `loadSalienceRankScores` outright (`src/indexer/search/ranking.ts`), along with the now-dead `maintenance-barrier`/`state-db` imports. `src/indexer/search/` has zero remaining `maintenance-barrier`/`state-db` dependency and opens no `state.db` handle on the search path (verified by grep).
- **Keeps** `salienceRankingContributor` and its `SALIENCE_WEIGHT` / `SALIENCE_MAX_BOOST` constants exported (now unwired) for a future gated, outcome-backed experiment.
- **Does not touch** `rank_score`'s computation, storage, or improve-internal consumption (`src/commands/improve/salience.ts` is untouched) — only its promotion into user-facing ranking is removed.
- `RankEntriesOptions.salienceRankScores` keeps its shape; its doc comment now says explicitly that `undefined`/`null` mean "no data" and a `Map` only matters to a caller that explicitly builds its own utility-contributor list including the contributor (there is no state.db fallback load anymore).

### Rationale (from the issue / 0.9.0-rc.5 triage)

- The boost is retrieval-dominated (`w_r = 0.60`, `src/commands/improve/salience.ts:356-360`) with no source filter, so it double-counts the same `usage_events` the utility-score contributor already reinforces.
- It warm-starts non-zero with no outcome evidence (`salience.ts:293-297`).
- `asset_salience` has zero pack coverage — the boost could only ever favor self-generated personal assets over an equally-relevant pack asset.
- It measures as noise on live data: max observed `rank_score` 0.356 → max multiplier ×1.071, mean ×1.016.
- Removing the default `state.db` load also fixes a confirmed hot-path defect: whenever `state.db` already existed, every default search synchronously called `acquireMaintenanceActivitySync("state-db")` (`src/core/maintenance-barrier.ts:91-107`) — a blocking `Atomics.wait` poll loop, up to 5s — plus a lock-file create, before the load's own 250ms SQLite `busy_timeout` ever applied.

## Test-first discipline (RED → GREEN, not squashed)

1. `2cb6d07` **test(search): pin salience-removal contract for #692 (RED)** — rewrote `tests/integration/ranking-salience-boost.test.ts` and `tests/integration/ranking-contributor-ablation.test.ts` against the *old* production code. Confirmed RED for the right reason (both files fail to import a symbol the implementation doesn't export yet):
   ```
   SyntaxError: Export named 'salienceRankingContributor' not found in module
   '.../src/indexer/search/ranking-contributors.ts'
   0 pass / 2 fail / 2 errors
   ```
2. `7a2ae78` **test(search): fix two test bugs found while confirming RED (still RED)** — fixed two bugs discovered while iterating (see commit body: an `akmIndex`-touches-`state.db` false assumption, and a wall-clock `expect(elapsed).toBeLessThan(...)` pattern this repo's own `scripts/lint-tests-isolation.ts` forbids as CI-flaky). Re-confirmed RED at this exact commit by stashing the (still-uncommitted) production fix and re-running — same import error, 0 pass / 2 fail.
3. `8dac8e3` **fix(search): remove the R2 salience boost from default ranking (#692)** — the production change. Turns the same two files GREEN: 21 pass / 0 fail.
4. `a30ffe5` **docs(changelog): note the R2 salience boost removal (#692)**.

## Golden-eval deltas

`tests/integration/curate-golden-eval.test.ts` run against the pre-fix code (git-checked-out parent commit, same test file) vs. post-fix (this branch), same deterministic-embedder fixture:

| metric | before (with boost) | after (this PR) | delta |
|---|---|---|---|
| meanScore | 0.890 | 0.890 | 0.000 |
| meanNdcg | 0.854 | 0.854 | 0.000 |
| meanRecall | 0.867 | 0.867 | 0.000 |
| meanMrr | 0.900 | 0.900 | 0.000 |
| meanNoBannedAboveRequired | 1.000 | 1.000 | 0.000 |
| totalBannedLeapfrog | 0 | 0 | 0 |

Every one of the 10 per-case rows (docker-homelab-deploy, release-coordination, testing-practices, git-workflow, code-review, ci-pipeline, deploy-production, seo-marketing, residue-docker, residue-release) is bit-for-bit identical before/after. **Exactly neutral, as expected** — the golden suite already floors below this value and this change removes a signal that measured as noise on live data. No rebaselining needed or done.

## Test plan

- [x] `bun test tests/integration/ranking-salience-boost.test.ts tests/integration/ranking-contributor-ablation.test.ts` → 21 pass, 0 fail (confirmed RED beforehand, see above)
- [x] Blast radius: `bun test tests/integration/curate-golden-eval.test.ts tests/integration/ranking-regression.test.ts tests/integration/graph-boost-ranking.test.ts tests/integration/scoped-utility.test.ts tests/integration/downstream-value-attribution.test.ts tests/ranking-capture-mode.test.ts tests/ranking-lesson-strength.test.ts tests/ranking-utility-decay.test.ts` (plus the two files above) → **111 pass, 0 fail across 10 files**
- [x] `bun run test:unit` → **2986 pass, 0 fail across 238 files**
- [x] `bun run test:integration` (full suite) → **5090 pass, 0 fail across 4 process-shards (380/380 files)**
- [x] `bun run lint` → exit 0 (pre-existing repo-wide `noNonNullAssertion` warnings only, unrelated to this change; `lint-tests-isolation`, `lint-runtime-boundary`, and every other custom lint script report OK)
- [x] `bunx tsc --noEmit` → clean, no output
- [x] Manually verified (scratch script, not committed) that `acquireMaintenanceActivitySync` really blocks ~5003ms under barrier contention before failing open, confirming the regression guard's premise

## Checklist

- [x] Tests pass (`bun test`)
- [x] Docs updated (`CHANGELOG.md` — `## [Unreleased]` → `### Changed`)

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -350,6 +350,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   installed scheduler entries are byte-identical to before, so upgrading shows no
   spurious drift.
 
+- **The R2 salience ranking boost no longer applies to default `search`/`curate`
+  ranking** (#692). `asset_salience.rank_score` (an encoding + outcome +
+  retrieval projection, recomputed every `improve` run) previously composed
+  into every default search as a bounded multiplicative boost
+  (`salience-ranking`, ×[1.0–1.2]), loaded best-effort from `state.db` on the
+  hot path. On live data it measured as noise (max observed multiplier
+  ×1.071, mean ×1.016): the boost was retrieval-dominated with no source
+  filter — double-counting the same `usage_events` the utility-score
+  contributor already reinforces — warm-started non-zero with no outcome
+  evidence, and had zero pack coverage, so it could only ever favor
+  self-generated personal assets over an equally-relevant pack asset.
+  Removing the default `state.db` load also fixes a confirmed hot-path
+  defect: whenever `state.db` already existed, every default search
+  synchronously waited on the maintenance-activity barrier before ranking
+  could even start — up to a 5-second stall on a blocking wait loop, plus a
+  lock-file create, before the load's own 250ms SQLite `busy_timeout` ever
+  applied. No config gate was added: a key for a term being removed would be
+  dead surface for the upcoming 1.0 contract freeze to carry forever.
+  `rank_score` itself, and everything `improve` computes and does with it
+  internally, are unchanged — only its promotion into user-facing ranking is
+  removed. The contributor stays in the codebase (unwired) for a future
+  gated, outcome-backed experiment.
+
 ### Fixed
 
 - **The compiled standalone binary can run `akm migrate`.** Release binaries
--- a/src/indexer/search/ranking-contributors.ts
+++ b/src/indexer/search/ranking-contributors.ts
@@ -60,13 +60,19 @@ const UTILITY_WEIGHT = 0.5;
 const UTILITY_MAX_BOOST = 1.5;
 
 /**
- * R2 — weight of the improve
- * loop's `asset_salience.rank_score` in user-facing ranking. Bounded well
- * below the utility boost so the composed signal refines, never dominates,
- * lexical/semantic relevance. rank_score ∈ [0,1] → boost ∈ [1, 1.2].
+ * R2 / #692 — weight of the improve loop's `asset_salience.rank_score`.
+ * rank_score ∈ [0,1] → boost ∈ [1, 1.2]. Bounded well below the utility boost
+ * so the composed signal would refine, never dominate, lexical/semantic
+ * relevance.
+ *
+ * As of #692 this is NOT wired into default user-facing ranking — see
+ * {@link salienceRankingContributor}'s doc comment. Kept exported (with the
+ * contributor) for a future gated experiment; do not reintroduce it into
+ * {@link defaultUtilityRankingContributors} without a measured
+ * curate-golden-bench delta and an explicit config gate.
  */
-const SALIENCE_WEIGHT = 0.2;
-const SALIENCE_MAX_BOOST = 1.2;
+export const SALIENCE_WEIGHT = 0.2;
+export const SALIENCE_MAX_BOOST = 1.2;
 
 /**
  * Phase 2A / Rec 5: default recency half-life (days) used when no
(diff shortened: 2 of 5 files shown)


itlackey:
Reviewed and merged (merge commit `03b3a30`). Verification record — everything below was run locally in a scratch clone at head `a30ffe5`, not taken from the PR body.

## The regression guard is real — proven by reintroduction

The PR replaced the wall-clock `expect(elapsed).toBeLessThan(...)` assertion (forbidden by `scripts/lint-tests-isolation.ts` Rule 3) with a 2500 ms test-level timeout as the promptness assertion. I verified this actually guards the stall, not just decorates it:

- **Experiment A** — restored pre-PR `src/indexer/search/ranking.ts` wholesale (i.e. `loadSalienceRankScores` + its `acquireMaintenanceActivitySync` call + the `undefined → load` fallback in `applyRankingRules`) on a scratch branch, keeping the new exports so no import error could mask behavior. Result: `search completes promptly even while the maintenance barrier is held` **fails** — `timed out after 2500ms`, actual runtime 5145 ms. Bun enforces the per-test timeout retroactively even though the body blocks synchronously in `Atomics.wait`, so the stall reports as a hard failure.
- **Experiment B** — full behavioral revert (also re-added `salienceRankingContributor` to `defaultUtilityRankingContributors`): **4 tests fail** — the timeout guard, `NOT applied via the default contributor list`, the one-entry-list pin, and the explicit-list ablation test. The removal contract is pinned behaviorally from every direction; this also compensates for the RED commits' import-error-shaped failure (weaker evidence on its own, since it proves the symbol was missing rather than the behavior wrong).

Two observations for future maintainers (non-blocking):
1. The `default search never creates state.db` test alone would **not** catch a loader reintroduction — the old loader short-circuited on a missing file and never created it (Experiment A left that test green). The 2500 ms budget test carries the guard alone.
2. Consequently, if that test's timeout is ever raised above ~5 s, the guard dies silently. The budget is load-bearing, not cosmetic.

## RED → GREEN confirmed

- `2cb6d07`: `SyntaxError: Export named 'salienceRankingContributor' not found` — 0 pass / 2 fail / 2 errors (verbatim as claimed)
- `7a2ae78`: same — 0 pass / 2 fail / 2 errors
- `8dac8e3`: 21 pass / 0 fail

## No test weakened

`git diff origin/main...HEAD -- tests/`: the ablation test was tightened (now pins `defaultUtilityRankingContributors` to exactly `["utility-ranking"]`, and proves the ablation mechanism against an explicitly supplied list, which the comment correctly notes avoids a vacuous assertion). The contributor math block (1.2×/1.1×/no-op bounds) is preserved via explicit injection — `applyUtilityContributors`' optional third parameter predates this PR, so no production signature was bent for the tests. The deleted assertions all targeted the deleted `loadSalienceRankScores` and died with it legitimately.

## Scope claims verified

- `grep -rn "maintenance-barrier|state-db|openStateDatabase|getStateDbPath" src/indexer/search/` → zero hits (remaining "state.db" strings are doc comments; `options.db` is the index DB)
- `salienceRankingContributor`, `SALIENCE_WEIGHT`, `SALIENCE_MAX_BOOST` all exported from `ranking-contributors.ts`
- `defaultUtilityRankingContributors` is exactly `[utilityRankingContributor]`
- `src/commands/improve/salience.ts`: empty diff vs main — untouched
- CHANGELOG entry sits under `## [Unreleased]` → `### Changed` and is accurate

## Checks I ran

- `bun run lint` → exit 0 (all custom lint scripts OK, including `lint-tests-isolation`)
- `bunx tsc --noEmit` → exit 0
- 9-file ranking blast radius (`ranking-salience-boost`, `ranking-contributor-ablation`, `curate-golden-eval`, `ranking-regression`, `graph-boost-ranking`, `scoped-utility`, `ranking-capture-mode`, `ranking-lesson-strength`, `ranking-utility-decay`) → **101 pass / 0 fail**
- CI: 5/5 check runs green (`check`, `node-smoke (22)`, `node-smoke (24)`, `slow-gated-tests`, `smoke`) — both `ci.yml` and `akm-eval-smoke.yml` fired as expected

## One honesty note on the golden-eval claim (non-blocking)

The reported exact 0.000 deltas are **structural, not measured**: `curate-golden-eval.test.ts`'s harness copies the frozen fixture, indexes, and curates — it never runs `improve` and never populates `asset_salience`, so the pre-PR "before" run's loader returned an empty map and the contributor was already a no-op. Bit-identical before/after was guaranteed by construction; the eval could not have shown a nonzero delta for this change. The PR body's "Exactly neutral, as expected — the golden suite already floors below this value" undersells this. It remains valid blast-radius evidence (a botched ranking refactor would have moved the numbers), and the removal contract is independently pinned by the dedicated tests above, so this does not block a removal — but the record should reflect that the golden eval was vacuous on the salience axis specifically.

This unblocks #672 (Workstream B), which relocates salience SQL that this PR's deletion of `loadSalienceRankScores` no longer duplicates.

---

Issue #644, opened 2026-06-22
improve: high-salience gate selects on type-weight stub, not content-derived encoding_salience (180/181 refs)


itlackey:
## Summary

The `akm improve` **high-salience reflect-eligibility gate** (#608) admits zero-feedback refs whose `encoding_salience >= salienceThreshold` (default 0.75). But for ~all assets that value is **not the content-derived score** from `scoreEncodingSalience` — it is the `DEFAULT_TYPE_ENCODING_WEIGHTS` **type stub**, re-asserted on every improve run. As a result the gate effectively means "reflect every feedback-less agent/command/skill/lesson," regardless of actual content value.

The per-run *churn* this caused was fixed in #643 (added a once-per-asset cooldown). **This issue tracks the deeper problem: the score the gate keys on is wrong.**

## Evidence (measured)

- **180 of 181** assets with `encoding_salience >= 0.75` carry exact type-weight values; only **4** rows in the entire stash (all `memory:`) have a real `scoreEncodingSalience` value.
  - Distribution of stored `encoding_salience`: `0.9` ×61 (agent/skill), `0.8` ×56 (command/workflow), `0.75` ×63 (lesson), `0.7` ×1539 (knowledge), `0.5` ×14 (memory) — these match `DEFAULT_TYPE_ENCODING_WEIGHTS` exactly.
- The two churning refs from the 24h health investigation:
  - `agent:…/community-manager` → stored `0.9`; content-formula would yield ~`0.65`.
  - `command:…/stack-log-correlation` → stored `0.8`; content-formula would yield ~`0.74` (i.e. *below* the 0.75 threshold).
  - Both had **zero** `show`/`feedback` events before they were flagged.

## Root cause

1. `scoreEncodingSalience` (`src/commands/improve/encoding-salience.ts`) — formula `novelty*0.40 + magnitude*0.35 + predictionError*0.25` — is called **only** from the distill path (`src/commands/improve/distill.ts:~943`) and persisted via `upsertAssetSalience` (`distill.ts:~957-973`). Distill has run for a tiny fraction of the stash, so almost no asset has a real score.
2. The improve run's salience loop (`src/commands/improve/improve.ts:~3378`) builds `computeSalience` inputs **without** `encodingSalience`. When absent, `computeSalience` falls back to `DEFAULT_TYPE_ENCODING_WEIGHTS[type]` (`src/commands/improve/salience.ts:~239-242`, table at `~119-128`).
[...]

Issue #499, opened 2026-05-26
recordImproveRun: capture distinct startedAt/completedAt timestamps for per-run wallTimeMs


itlackey:
## Symptom

`improve_runs.started_at` and `improve_runs.completed_at` are written to the same end-of-run timestamp by `recordImproveRun()` (in `src/core/state-db.ts`). The CLI's per-run summary (`akm health --detail per-run`) gets `wallTimeMs: 0` when computed from those fields directly. The current fix joins to `task_history` via interval containment (063e156) and works for cron-driven runs, but:

- Manual `akm improve` invocations (no task_history wrapper) still report `wallTimeMs: 0` per-run because no enclosing task interval exists.
- The join is an indirect proxy — would be cleaner to capture real start/end on the row itself.

## Root cause

`recordImproveRun()` is called at run END. Both `startedAt` and `completedAt` parameters are populated from the same `endTimestamp` by every caller. The function's signature accepts distinct values, but no caller passes them.

```ts
// Current callers (improve.ts):
recordImproveRun(db, {
  id, startedAt: endTimestamp, completedAt: endTimestamp, ...
});
```

## Proposal

1. In `src/commands/improve.ts`, capture the run's actual start timestamp at the top of `akmImprove()` (before any work begins).
2. Thread it through to the `recordImproveRun()` call so `startedAt = realStart` and `completedAt = realEnd`.
3. Update `buildPerRunSummaries` in `src/commands/health.ts` to prefer the row's own delta when both timestamps differ, falling back to the task_history join only when they're equal (the current legacy fallback).

## Acceptance

- [ ] Manual `akm improve` produces an `improve_runs` row with `started_at < completed_at`
- [ ] `akm health --detail per-run` reports a non-zero `wallTimeMs` for those manual runs without needing task_history
- [ ] Cron-driven runs continue to report wall times — either from the row's own delta or the task_history join (either is correct)
- [ ] Schema-version bump on `improve_runs` is NOT required — the columns already exist, only the values change
- [ ] Existing backfill data (timestamps equal) keeps falling back to task_history correctly

## References

- Phase 1: `feat(health): richer improve metrics from improve_runs` (commit 379e5c4) — wallTime block introduced
- Phase 2 wallTime bug: per-run wallTime fell back to 0 because of identical start/end (commit 063e156 fixed via task_history join)
- 2026-05-25 incident notes: per-run wallTimeMs initially 0 across all rows; interval-containment join fixed the cron case but manual invocation case remains

[...]

Issue #733, opened 2026-07-30
improve: orphan-GC pass for state rows whose refs no longer resolve


itlackey:
## Problem

When an asset is renamed, deleted, or migrated, its `asset_salience`, `asset_outcome`, and usage-event rows stay behind keyed by the old ref. This is a demonstrated field-failure class: the 2026-07-28 environment repair found 30 salience + 30 outcome rows stranded under a retired source root (`~/code/dimm-city/agent-stash//…`) plus stale usage refs from reserved-file renames, all requiring manual SQL to fix.

With `akm mv` removed in the 0.9 CLI overhaul (renames are delete + create per STABILITY), plain filesystem renames become the normal path, so orphaned rows will accumulate steadily rather than exceptionally. The opt-in `scripts/rekey-asset-ref.ts` helper preserves identity when a user chooses to, but hygiene should not depend on users remembering to run it.

## Proposal

An orphan-GC pass in the improve loop (retention/maintenance lane) that drops state rows whose asset refs no longer resolve in the index.

## Design constraints to scope

- **Absent ≠ deleted.** A source that fails to scan preserves last-known-good index rows; GC must key off authoritative deletion (asset absent after a *complete* scan of its source), never off temporary unavailability of a bundle, or it will destroy signal for read-only/cache-backed sources that are briefly unreachable.
- **Grace window** before collecting, mirroring the txn sweeper pattern, so a rename followed promptly by `rekey-asset-ref.ts` doesn't race the GC.
- **Prove before deleting**: first ship as a counter/report (health advisory or improve telemetry: N orphaned rows across M refs), then enable deletion — consistent with the prove-or-delete approach used for ranking metrics.
- **Emit events** for what was collected so the trail survives in `akm log`.
- Interaction with #692 (salience ranking value unproven): if salience is removed from default ranking, the GC scope shrinks accordingly.

## Origin

Split out of the 0.9 CLI overhaul (`feature/cli-overhaul-0.9`): the mv drop + re-key script landed there; this pass touches the improve loop and is deliberately scoped separately.

itlackey:
**Moved 0.9.0 → 0.9.1** during milestone re-triage: 0.9.0 is at rc.11 in close-out, and scoping this against the tree shows it is not an rc-sized change. Three pieces of plumbing the design constraints require do not exist yet:

[...]

PR #737, opened 2026-07-30
refactor(state): extract salience/outcome repositories and guard state-table SQL (#672)

### LINKED ISSUE #672: Architecture follow-up: extract salience/outcome repositories and widen the X4 guard

itlackey:
## Summary

Part 2 of #672 (part 1 shipped at `028cc3de`). Extracts the raw `asset_salience` /
`asset_outcome` SQL out of `src/commands/improve/{salience,outcome-loop}.ts` into
new `src/storage/repositories/{salience,outcome}-repository.ts` files, mirroring
the existing state.db repository precedents (`proposals-repository.ts`,
`improve-runs-repository.ts`, `events-repository.ts`): queries and row-mapping
extracted verbatim, only relocated behind the repository boundary. A new
`getTopRetrievalSalience(db, limit)` repository entry point absorbs the health
read in `src/commands/health/metrics.ts` (previously an inline, hardcoded
`LIMIT 100`). `salience.ts` and `outcome-loop.ts` re-export the moved symbols so
existing importers and tests resolve unchanged. A new third `state-table-sql`
rule is added to `scripts/lint-repository-sql.ts` guarding raw SQL naming these
two tables everywhere under `src/` except the repository directory and
`src/core/state/migrations.ts`.

**Two corrections to the issue text**, discovered against the tree (also called
out in `docs/architecture/specs/0.9.0-close-out-plan.md`, Workstream B):

1. **Location.** The issue proposes `src/core/state/salience-repo.ts` /
   `outcome-repo.ts`. The tree's actual convention is
   `src/storage/repositories/*-repository.ts` (22 files at plan-authoring time,
   24 now with this PR's 2 additions — including the state.db precedents
   named above). This PR follows the tree, not the issue.
2. **Guard mechanism.** The issue calls the widening "a one-line allowlist
   change" (add `improve/**` to the X4 guard's `GUARDED_PREFIXES`). That's
   incorrect: the existing two rules (`db-owner-import`, `db-open-call`)
   detect DB-owner imports and direct database opens — not raw SQL.
   `salience.ts` / `outcome-loop.ts` take a `db` parameter and import no
   DB-owner module, so they already pass both existing rules today with their
   SQL fully intact; widening the prefix list would enforce nothing against
   them. A third rule that inspects raw SQL text was required instead
   (`state-table-sql`, reusing the existing comment/string-stripping
   machinery with `keepStrings: true` so it can see SQL text and import
   specifiers, the same way string content is inspected today).

### RED → GREEN commit order for the lint rule

1. `63df6c6` **test(lint)** — add `state-table-sql` fixtures to
   `tests/integration/lint-repository-sql.test.ts` (RED: 8 pass / 3 fail,
   `ruleId "state-table-sql" not found`).
2. `a55a3ff` **feat(lint)** — implement the rule itself, with per-rule scoping
   (`Rule.appliesTo`) since its scope (everywhere under `src/` except the
   repository dir and `migrations.ts`) is fundamentally different from the
   other two rules' narrow `GUARDED_PREFIXES` (GREEN: 11 pass / 0 fail).

### The import-cycle fix (commit `a8a0471`)

The first refactor commit (`d90eddd`) introduced a real regression that
`bun run test:unit` caught: `tests/architecture/import-cycle-ratchet.test.ts`
enforces a zero-tolerance, shrink-only baseline (`CYCLE_PARTICIPANT_BASELINE`
is empty) over the whole `src/` static import graph, and it counts type-only
imports as graph edges the same as value imports. The initial repository
files each imported a type back from the domain module they re-export from —
`salience-repository.ts` imported `EncodingSource`/`SalienceVector` from
`commands/improve/salience.ts`, and `outcome-repository.ts` imported
`AssetOutcomeRow` from `commands/improve/outcome-loop.ts` — while the domain
modules import the repositories' functions for re-export. Two 2-node cycles.
The baseline can't be widened to admit them (the ratchet is deliberately
absolute).

**Fix:** make both repository files import nothing from the domain modules
they serve, so the mandatory domain→repository edge (for re-export) stays
one-directional:

- `salience-repository.ts`: `AssetSalienceRow.encoding_source` now uses the
  inline literal union `"content" | "type-stub" | null` (structurally
  identical to `EncodingSource`) instead of importing the type; a local
  `SalienceVectorLike` interface (listing only the fields
  `upsertAssetSalience` reads) replaces the `SalienceVector` import — any
  real `SalienceVector` value is still structurally assignable, so callers
  are unaffected.
- `outcome-repository.ts`: `AssetOutcomeRow` (plain string/number fields, no
  cross-type dependency) now lives here instead of in `outcome-loop.ts`,
  which imports it back and re-exports it — mirroring the `AssetSalienceRow`
  precedent (already living in its repository) exactly.

**Consequence — the one real scope change worth flagging honestly:**
`AssetOutcomeRow`'s canonical *definition* moved from `outcome-loop.ts` into
`outcome-repository.ts` (the same place `AssetSalienceRow` was already defined
from the first commit). This is an internal detail, not a contract change: I
independently verified (grep across `src/` and `tests/`, not just trusting the
commit message) that **no importer or test import path changed** — 6 improve
importers of `salience.ts` (`collapse-detector.ts`, `distill.ts`,
`distill/quality-gate.ts`, `loop-stages.ts`, `preparation.ts`,
`proactive-maintenance.ts`), 9 test files importing `.../improve/salience` and
4 importing `.../improve/outcome-loop` (12 unique files — consistent with the
commits' "~11 dependent test files"), and zero test files import the new
`storage/repositories/*-repository` paths directly except the lint meta-test
itself (which references them as string literals, not imports). Both new type
declarations are pure `interface`/type-alias shapes with no runtime
initialization, so removing the import doesn't introduce a
partially-initialized-module hazard — there is now no cycle at all between
these four files (independently confirmed: `bun scripts/lint-import-cycles.ts`
→ "0 cycle participant(s)", and `bun test
tests/architecture/import-cycle-ratchet.test.ts` → 4 pass / 0 fail).

Behavior is unchanged by the fix: both changes are pure type-shape
adjustments, not logic changes. The CASE guards, SQL text, and bind order
from the first commit are untouched (independently re-verified below, not
just taken on the commit messages' word).

### Verbatim-move spot checks (re-diffed directly against the pre-refactor originals)

- `upsertAssetSalience`'s two `#644` `CASE WHEN asset_salience.encoding_source
  = 'content' AND excluded.encoding_source = 'type-stub' …` guards (preserving
  `encoding_salience` / `encoding_source` provenance) are present in
  `salience-repository.ts`, byte-for-byte identical to the pre-refactor SQL in
  `commands/improve/salience.ts`, including bind order.
- `upsertAssetOutcome`'s `review_pressure` column is (still) omitted from both
  the `INSERT` column list and the `ON CONFLICT ... DO UPDATE SET` clause in
  `outcome-repository.ts`, byte-for-byte identical to the pre-refactor SQL,
  with the same doc-comment rationale (`DEFAULT 0` seeds fresh rows; omitting
  it from `SET` leaves an existing row's value untouched going forward).
- `getTopRetrievalSalience(db, limit)` is genuinely new (not a move): the
  pre-refactor `commands/health/metrics.ts` had the identical query
  hardcoded at `LIMIT 100`; the repository version parameterises `limit` and
  the call site passes `100`, preserving the existing fail-open try/catch.
- Confirmed Workstream A (#692, already merged to `main` at `03b3a305`)
  already removed the only other raw-SQL site the plan flagged
  (`ranking.ts`'s `loadSalienceRankScores`) — nothing left for this PR to
  relocate there.

## Test plan

**Verification was performed post-hoc by a second agent**, after the original
implementer's container was suspended mid-task (before it could verify, push,
or open a PR). The 5 commits were reviewed against the diff and the close-out
plan, then the following was actually run against this branch:

- `bun run lint` — **exit 0**. `lint-repository-sql: OK — … no raw
  asset_salience/asset_outcome SQL lives outside it`. ~1459 pre-existing biome
  `noNonNullAssertion` warnings are present but unrelated to this PR (verified
  none land in any file this PR touches; `bun run lint`'s exit code, not
  warning count, gates the script).
- **Proved the new rule actually bites**, rather than trusting a zero count:
  added a scratch file under `src/commands/` containing raw
  `db.prepare("SELECT * FROM asset_salience WHERE asset_ref = ?")`, ran
  `bun scripts/lint-repository-sql.ts` directly → 1 violation,
  `[state-table-sql]`, exit 1. Removed the scratch file → back to OK, exit 0,
  clean tree confirmed via `git status`.
- `bunx tsc --noEmit` — clean, no output, exit 0.
- `bun run test:unit` — **2986 pass / 0 fail** (238/238 files, 4 shards) —
  matches the implementer's commit-message claim exactly.
- The refactor's safety-net suites, run explicitly together — **164 pass / 0
  fail** (370 `expect()` calls, 9 files):
  `tests/integration/commands/improve/{salience,salience-wiring,outcome-loop,outcome-loop-wiring,outcome-invariance}.test.ts`,
  `tests/integration/health-ws5-observability.test.ts`,
  `tests/commands/improve/encoding-salience.test.ts` (note: this lives at
  `tests/commands/improve/encoding-salience.test.ts`, not
  `tests/encoding-salience.test.ts`),
  `tests/integration/ranking-salience-boost.test.ts`,
  `tests/integration/lint-repository-sql.test.ts`.
- `git diff origin/main...HEAD -- tests/` — confirmed the **only** test file
  touched is `tests/integration/lint-repository-sql.test.ts` (new
  `state-table-sql` fixtures + the renamed assertion that
  `commands/improve/preparation.ts` is unguarded *for the first two rules but
  not the third*). No other pre-existing test was edited — verified directly,
  not assumed.
- `bun run test:integration` (full suite, all 380 files, 4 shards): **5093
  pass / 2 fail** / 13 skip (6422 `expect()` calls). Both failures are
  **pre-existing host-contention flakes, unrelated to this PR**:
  `tests/integration/three-db-cutover.test.ts` ("resumes after crashing
  immediately after current proposal repair commit") and
  `tests/integration/workflows/native-executor.test.ts` ("the lifetime unit
  cap is seeded from the run's journal") timed out at 30s. Neither file is
  touched by this PR's diff (`git diff origin/main...HEAD --stat` confirms
  both are absent). A second full `test:integration` run was independently
  executing concurrently on the same host for the duration of this run (a
  sibling agent's worktree, confirmed via `ps`, whose own shard included both
  of these exact files at the same time) — matching this project's own
  documented host-load-flake pattern (see the `#499` note referenced in
  `docs/architecture/specs/0.9.0-close-out-plan.md`). I re-ran both failing
  files in isolation immediately after: both are **clean** —
  `three-db-cutover.test.ts` 32 pass / 0 fail, `native-executor.test.ts` 66
  pass / 0 fail. Total pass+fail count (5095) matches the original
  implementer's commit-message claim (`5095 pass / 0 fail`) exactly; only the
  pass/fail split differs, consistent with a load-induced flake rather than a
  regression.
- Independently re-verified `bun scripts/lint-import-cycles.ts` (→ 0
  participants) and `bun test tests/architecture/import-cycle-ratchet.test.ts`
  (→ 4 pass / 0 fail) rather than relying on the commit message's claim.

## Checklist

- [x] Tests pass (`bun test`)
- [ ] Docs updated (if applicable) — n/a, internal-only refactor;
      `docs/architecture/specs/0.9.0-close-out-plan.md` (Workstream B) already
      documents the intended shape and is unchanged by this PR. CHANGELOG
      updated under `### Changed`, placed next to the sibling #692 entry per
      existing precedent for internal-only bullets.

Fixes #672

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -373,6 +373,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   removed. The contributor stays in the codebase (unwired) for a future
   gated, outcome-backed experiment.
 
+- **Internal: `asset_salience` / `asset_outcome` state.db access moved behind
+  `src/storage/repositories/{salience,outcome}-repository.ts`** (#672 part 2).
+  Mirrors the existing state.db repository precedents
+  (`proposals-repository.ts`, `improve-runs-repository.ts`,
+  `events-repository.ts`): the raw SQL, row-mapping, and the #644
+  encoding-provenance CASE guards are extracted verbatim, only relocated —
+  `commands/improve/salience.ts` and `outcome-loop.ts` re-export the moved
+  functions, so no importer or test churns. A new `state-table-sql` rule in
+  `scripts/lint-repository-sql.ts` now fails the build if raw
+  `asset_salience`/`asset_outcome` SQL reappears outside the repository
+  directory (or `core/state/migrations.ts`). Not a user-visible behavior
+  change: `rank_score`, `outcome_score`, and everything `improve`/`health`
+  compute from them are identical.
+
 ### Fixed
 
 - **The compiled standalone binary can run `akm migrate`.** Release binaries
--- a/scripts/lint-repository-sql.ts
+++ b/scripts/lint-repository-sql.ts
@@ -17,15 +17,33 @@
  * D5 (WorkflowDocuments reader) removed those inversions; this guard ratchets
  * them shut so they cannot regrow.
  *
- * Scope is intentionally NARROW. Raw SQL legitimately lives in many other
- * modules (health, indexer, usage-telemetry, the improve read-side, …) and the
- * codebase does NOT funnel all SQL through one repository directory — so a
- * blanket "no SQL outside repositories" rule would be a large status-quo
- * allowlist, not a real boundary. This guard enforces only the boundary that
- * was actually inverted and actually regressed.
+ * Scope for `db-owner-import` / `db-open-call` is intentionally NARROW. Raw
+ * SQL legitimately lives in many other modules (health, indexer,
+ * usage-telemetry, the improve read-side, …) and the codebase does NOT funnel
+ * all SQL through one repository directory — so a blanket "no SQL outside
+ * repositories" rule would be a large status-quo allowlist, not a real
+ * boundary. Those two rules enforce only the boundary that was actually
+ * inverted and actually regressed.
  *
- * Comments and string literals are stripped before matching, so prose that
- * merely mentions these names does not trip the guard — only real code does.
+ * A third rule, `state-table-sql` (#672 part 2), is scoped differently and
+ * much more broadly: raw SQL naming the `asset_salience` / `asset_outcome`
+ * state.db tables is guarded EVERYWHERE under `src/` except
+ * `src/storage/repositories/**` (their new home,
+ * salience-repository.ts / outcome-repository.ts) and
+ * `src/core/state/migrations.ts` (schema DDL, not application read/write
+ * SQL — migrations are append-only history and are never refactored to use
+ * a repository). Unlike the first two rules, a widened `GUARDED_PREFIXES`
+ * would not express this: `salience.ts` and `outcome-loop.ts` take `db`
+ * parameters and import no DB-owner module, so they pass the first two rules
+ * today with their SQL fully intact — unlike `db-owner-import`/`db-open-call`,
+ * this is a raw-SQL-content rule, not an import/open-call rule, so it needs
+ * its own pattern and its own scope predicate (`isStateTableSqlScope` below),
+ * not just a wider prefix list.
+ *
+ * Comments and string literals are stripped before matching (except for rules
+ * that opt into `keepStrings`, which need to see string content — import
+ * specifiers and SQL text both live inside string literals), so prose that
+ * merely mentions these names does not trip any rule — only real code does.
  *
  * Exit codes: 0 — clean; 1 — violations (or internal error).
  * Usage: bun scripts/lint-repository-sql.ts
@@ -41,15 +59,49 @@ const srcDir = path.join(repoRoot, "src");
  * Subsystems that must use repositories rather than DB internals. POSIX-relative
  * directory prefixes. Add a new boundary here only when its inversions are
  * already cleared (the guard must stay at 0 violations).
+ *
+ * Scope for `db-owner-import` and `db-open-call` ONLY — the `state-table-sql`
+ * rule (#672 part 2) has its own, much broader scope; see
+ * `isStateTableSqlScope` below. Left unchanged by #672 part 2 on purpose.
  */
 const GUARDED_PREFIXES: readonly string[] = ["src/registry/", "src/workflows/runtime/"];
+
+/**
+ * `state-table-sql` scope (#672 part 2): applies everywhere under `src/`
+ * except these directory prefixes …
+ */
+const STATE_TABLE_SQL_EXCLUDED_PREFIXES: readonly string[] = ["src/storage/repositories/"];
+
+/** … and except these exact files (POSIX-relative to repo root). */
+const STATE_TABLE_SQL_EXCLUDED_FILES: readonly string[] = ["src/core/state/migrations.ts"];
+
+/** Scope predicate for the original two rules: registry + workflow-runtime only. */
+function isGuardedPrefix(rel: string): boolean {
+  return GUARDED_PREFIXES.some((p) => rel.startsWith(p));
+}
+
+/**
+ * Scope predicate for `state-table-sql`: everywhere under `src/` except the
+ * repository directory (the new home for this SQL) and the migrations file
+ * (schema DDL, append-only, never behind a repository).
+ */
+function isStateTableSqlScope(rel: string): boolean {
+  if (!rel.startsWith("src/")) return false;
+  if (STATE_TABLE_SQL_EXCLUDED_FILES.includes(rel)) return false;
+  return !STATE_TABLE_SQL_EXCLUDED_PREFIXES.some((p) => rel.startsWith(p));
+}
 
 interface Rule {
   id: string;
   pattern: RegExp;
   message: string;
-  /** Match against string-preserving text (for import-specifier rules). */
+  /** Match against string-preserving text (for import-specifier / SQL-text rules). */
   keepStrings?: boolean;
+  /**
+   * Which files this rule applies to, given a repo-relative POSIX path.
+   * Defaults to `isGuardedPrefix` (the original two rules' scope) when absent.
+   */
+  appliesTo?: (rel: string) => boolean;
 }
 
 const RULES: readonly Rule[] = [
@@ -67,6 +119,15 @@ const RULES: readonly Rule[] = [
       /\b(?:openExistingDatabase|openIndexDatabase|openStateDatabase|openManagedDatabase)\s*\(|\bnew\s+Database\s*\(/,
     message:
       "opens a database directly — registry/workflow-runtime code must query through a repository in src/storage/repositories/**",
+  },
+  {
+    id: "state-table-sql",
+    // Raw SQL (or any code) naming the asset_salience / asset_outcome tables.
+    pattern: /\basset_salience\b|\basset_outcome\b/,
+    message:
+      "raw SQL names a state-table (asset_salience/asset_outcome) outside src/storage/repositories/** — move the query behind salience-repository.ts / outcome-repository.ts (repository-owns-SQL boundary, #672)",
+    keepStrings: true,
+    appliesTo: isStateTableSqlScope,
   },
 ];
 
@@ -162,7 +223,11 @@ interface Violation {
 
 /** Pure matcher: lint one file's content given its repo-relative POSIX path. */
 export function lintContent(rel: string, raw: string): Violation[] {
-  if (!GUARDED_PREFIXES.some((p) => rel.startsWith(p))) return [];
+  // Each rule has its own scope (appliesTo, default isGuardedPrefix). Skip the
+  // (relatively expensive) comment/string stripping entirely when NO rule
+  // could possibly apply to this file.
+  const applicableRules = RULES.filter((rule) => (rule.appliesTo ?? isGuardedPrefix)(rel));
+  if (applicableRules.length === 0) return [];
 
   const noStrings = stripCommentsAndStrings(raw, false).split("\n");
   const withStrings = stripCommentsAndStrings(raw, true).split("\n");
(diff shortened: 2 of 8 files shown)


itlackey:
Merged as reviewer of record after independent verification in a fresh clone at head `01212bb6` (base `03b3a305` = current `main`). Every load-bearing claim in the PR body was re-checked, not taken on faith. Findings: none blocking, none advisory — the PR is as described.

**Import-cycle fix (`a8a0471`) — verdict: the cycle was real and the fix is correct and precedent-consistent.**
- At `d90eddd`, `bun test tests/architecture/import-cycle-ratchet.test.ts` fails exactly as claimed: 4 files joined cycles (`salience.ts`, `outcome-loop.ts`, both repositories). I confirmed the two type-only back-imports existed in that commit. At head: 4 pass / 0 fail, and `bun scripts/lint-import-cycles.ts` → 0 participants.
- `SalienceVectorLike` is field-for-field identical to `SalienceVector` (`encoding`/`outcome`/`retrieval`/`rankScore` required, `encodingSource?: "content" | "type-stub"` optional) — not a narrowing; every field the SQL binds is present, so no latent bind gap.
- `AssetOutcomeRow` relocation: grepped `src/` + `tests/` myself. The only external consumer (`tests/integration/commands/improve/monitor-liveness.test.ts`) imports it from `outcome-loop`, which still re-exports it; the PR diff touches no importer file, so no import path changed. Re-exporting a type whose definition moved into a repository matches the `a66990b1` precedent (`core/state-db.ts` became a barrel of `export *` from the nine repositories) and the `AssetSalienceRow` handling in this PR.

**Lint rule — verdict: real teeth, correctly scoped, genuinely wired.**
- Reproduced the bite experiment myself: scratch file under `src/commands/improve/` with `db.prepare("SELECT * FROM asset_salience WHERE asset_ref = ?")` → 1 violation, `[state-table-sql]`, exit 1 from `bun scripts/lint-repository-sql.ts`, and — going one step further than the PR — full `bun run lint` with that file in place also fails at the `lint-repository-sql` step, proving end-to-end wiring through the `&&` chain, not mere script presence.
- Boundary probes, all correctly silent (exit 0): the identical SQL inside `src/storage/repositories/`; comment/prose-only mentions of the table names; and `usage_events` SQL under `src/commands/` (the plan's deliberate scope cut).
- RED→GREEN order verified empirically: meta-test at `63df6c6` → 8 pass / 3 fail; at `a55a3ff` → 11 pass / 0 fail.

**Verbatim-move check — byte-identical, mechanically diffed against `origin/main`:** `upsertAssetSalience`'s full body including both `#644` `CASE WHEN … encoding_source` guards and the 7-value bind order; all salience readers/plasticity helpers; the `asset_outcome` upsert SQL with `review_pressure` omitted from both the INSERT column list and the `ON CONFLICT … DO UPDATE SET` clause, bind order preserved through the values object; all outcome readers. `metrics.ts` now calls `getTopRetrievalSalience(db, 100)` inside the unchanged fail-open try/catch (`LIMIT ?` bound to 100 ≡ the old hardcoded `LIMIT 100`).

**Test hygiene:** `git diff origin/main...HEAD -- tests/` touches only `tests/integration/lint-repository-sql.test.ts`; the pre-existing `preparation.ts`-is-unguarded case keeps its original assertion (narrowed to the first two rules) and adds the state-table-sql positive — strengthened, not weakened. The PR's path correction is right: `tests/commands/improve/encoding-salience.test.ts` exists; `tests/encoding-salience.test.ts` does not.

**Commands run here (all green):** `bun run lint` → exit 0 · `bunx tsc --noEmit` → exit 0 · `bun run test:unit` → 2986 pass / 0 fail (238/238 files) · the 9 safety-net suites → 164 pass / 0 fail (370 expects) · ratchet test at head → 4/0 · `tests/integration/three-db-cutover.test.ts` in isolation → 32 pass / 0 fail · `tests/integration/workflows/native-executor.test.ts` in isolation → 66 pass / 0 fail.

**The two reported integration failures — determination: environmental, not blocking.** Neither file is in this PR's diff; both pass in isolation on my host; and CI's `check` job (`bun run check` = lint + tsc + test:unit + full `test:integration`) ran the entire 380-file suite green on this exact head. I could not retroactively observe the original host contention, so I state precisely: I did not reproduce the failure, and a quiet-host full-suite pass (CI) plus clean isolated runs rule out a code regression.

**CI at merge time:** all 5 checks success (`check`, `smoke`, `node-smoke (22)`, `node-smoke (24)`, `slow-gated-tests`); `mergeable_state` `clean`. Merge commit: `b0c7bea3ae238ed772e92636deb8d9257526f689`. Workstream C (#733) is unblocked.

---

PR #738, opened 2026-07-30
feat(okf): OKF v0.2 read support and proposal provenance stamping (#730)

### LINKED ISSUE #730: OKF v0.2 update
Update the OKF adapter to use the new spec, update AKM's spec to include these new front matter fields and akm processes to leverage the sources, generated, and other new frontmatter that will help track autonomous updates and verification using the proposals system

https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md

itlackey:
## Summary

Implements #730 (OKF v0.2) per Workstream D of
`docs/architecture/specs/0.9.0-close-out-plan.md`: read-side support for the
OKF v0.2 trust/provenance/lifecycle frontmatter families, provenance stamping
on AKM-native writes made through the proposal system, and the matching
conformance/spec documentation. **The `okf` adapter stays consumer-only** —
this was the plan's central scope decision and it holds (evidence below).

**Update (this revision):** the owner reviewed the two open naming/shape
questions the first revision flagged and made two decisions, plus the review
surfaced six small follow-up fixes. Both decisions and all six follow-ups are
applied in this revision — see "Owner decisions applied" and "Review
follow-ups addressed" below. This revision also rebases onto `main` (PR #737,
the salience/outcome repository extraction, merged after this PR was first
opened) and re-verifies everything from scratch.

### D1 — read side (RED `9812af9c` → GREEN `73c20c2e`)

`src/core/adapter/adapters/okf-adapter.ts`'s `recognize()` parses, all
leniently (never rejecting a document for a missing/malformed field):

- `updated` ← `generated.at`, taking precedence over the legacy `timestamp`,
  which remains a fully valid fallback (the v0.2 breaking-with-fallback
  change).
- `generated` (`by`/`at`), `verified` (list, or v0.2's single-mapping
  shorthand), `sources` (object list), `status`
  (`draft`/`stable`/`deprecated`), `stale_after`, and `okf_version`.

**Field-naming decision (D1.3):** AKM already has three fields at adjacent
names — `sources?: string[]` (wiki citation strings), `generation?: number`
(consolidation depth), and the `quality: "generated"` enum value — so v0.2's
families land under new, namespaced `IndexDocument` fields instead of
overloading them: `provenance?: OkfProvenance` (`{ generatedBy?, generatedAt?,
verified?, sources? }`), plus flat `lifecycleStatus?`, `staleAfter?`,
`okfVersion?` (`src/core/adapter/types.ts`). **The owner endorsed this
TypeScript shape as-is** — nothing about it changes in this revision, only
the on-disk frontmatter spelling `promoteProposal` writes (D2, below).

### D2 — write side (RED `61682fe3` → GREEN `23ccdbde`, shape revised `0e13e289`)

`promoteProposalWithLease` (`src/commands/proposal/repository.ts`) stamps
OKF v0.2 provenance onto every AKM-native markdown asset it promotes,
projecting bookkeeping the proposals system already tracked in `state.db`
but never wrote to disk:

- `generated.by` — keyed on **what produced the content**:
  `akm/<version>` for an automated source (reflect/distill/consolidate/
  extract/improve/schema-repair), `human:<id>` otherwise.
- `verified[].by` — keyed on **what accepted this promotion**:
  `akm/<version>` when a `gateDecision` was supplied (automated
  drain/triage), `human:<id>` for a direct `akm proposal accept`. Every
  promotion is exactly one of these two paths, so `verified` is always
  stamped, and now **accumulates** across re-promotions instead of the
  newest entry replacing the record.
- `evidenceSources`, when present, projects as `provenance.sources`
  (`{resource}` minimum) alongside the original key (not a rename).
- `DOCUMENT_JSON_CARRIED_FIELDS` (`akm-adapter.ts`) carries `provenance`, and
  `doc-to-entry.ts` unpacks it to a first-class `entry.provenance`, so `akm`'s
  own adapter rereads what it wrote and `akm show` surfaces it.

**On-disk shape — Decision 1 (owner review, applied this revision):** the
first revision nested everything (`generated`/`verified`/`sources`) under one
`provenance:` key. **The owner decided on a hybrid shape instead:**
`generated: {by, at}` and `verified: [{by, at}]` are now written **bare at
the top level**, exactly as OKF v0.2 spells them — matching `okf-support.md`'s
own "AKM Markdown is an OKF-compatible superset" claim for trust metadata,
since neither bare key has any pre-existing AKM consumer and nesting them
bought nothing. Only `sources` stays namespaced (`provenance: {sources}`),
because a bare top-level `sources:` on a promoted wiki page genuinely
collides with the pre-existing AKM-native wiki citation-**string** convention
(`applyWikiFrontmatter`, which silently drops non-strings). Stamped
frontmatter is durable on-disk state in user stashes, so this was worth
settling now rather than after assets accumulate under the old shape.

**Round-trips, verified:** `applyProvenanceFrontmatter`
(`src/indexer/passes/metadata.ts`) now reads the bare `generated`/`verified`
families plus namespaced `provenance.sources` into the same
`IndexDocument.provenance` shape (and still accepts the older fully-nested
spelling as a fallback, so nothing already stamped under the first revision
breaks). `tests/integration/proposals.test.ts`'s "the promoted asset
re-indexes with provenance surfaced on the persisted entry" test exercises
the full loop — `akmProposalAccept` writes, `akmIndex` + `indexedEntry` reads
back — and asserts `entry.provenance.generatedBy` / `.verified` are
populated. Green (see Test plan). The `okf` adapter's own read side (third-
party OKF v0.2 bundles) was already parsing the bare spelling from D1 and is
unaffected; both adapters land AKM-native and OKF-bundle documents in the
same `IndexDocument.provenance` shape.

**Workflow is now stamped — Decision 2 (owner review, applied this
revision):** the first revision's own PR body overstated the closed-
frontmatter-allowlist claim — asserting "workflow; command/agent/skill share
the posture." **The review corrected this: `workflow` is the ONLY asset type
with a closed frontmatter allowlist** (`src/workflows/validator.ts`);
command/agent/skill get only the generic + quality validators and tolerate
extra keys, same as knowledge/lesson/memory/fact/session/instruction. That
made `workflow` the sole type that silently promoted **unstamped** via the
re-validate-and-fall-back safety net — an undocumented coverage hole, not a
crash. Fixed by adding `generated`/`verified`/`provenance` to
`ALLOWED_FRONTMATTER_KEYS`, so workflow promotions are now stamped like every
other AKM-native markdown type. The re-validate-and-fall-back path stays as a
backstop (for a hypothetical future closed-allowlist type) but no longer
triggers for workflow — pinned by a new test,
`"workflow — the one closed-allowlist type — is stamped, not silently
skipped"`, which I verified actually depends on the fix (removing the three
keys from the allowlist makes it fail with `generatedBy` undefined; restoring
them makes it pass). The overstated inline comment in `repository.ts` is
corrected to match.

**Consumer-only scope — verified, not just asserted:**
- Stamping happens in `promoteProposalWithLease` *after*
  `assertAkmAssetWrite(mutationTarget.source)`, which throws before any
  stamping code runs whenever the target's adapter isn't `akm`. `git diff
  origin/main...HEAD -- src/core/write-source.ts` is **empty** — the
  OKF-rejection mechanism itself is untouched by this PR.
- `tests/integration/okf-conformance.test.ts`'s write-rejection block is
  untouched — confirmed green (see Test plan).
- Only `.md` targets are stamped; only markdown promotions reach the
  stamping code at all.

### D3 — conformance & docs (`21476f22`, extended this revision)

- Runbook renamed `okf-v0.1-conformance-runbook.md` →
  `okf-v0.2-conformance-runbook.md`; Rules 10-14 for the v0.2 families.
  `KC_REF` independently re-verified against the pinned upstream commit.
- `akm-0.9.0-bundle-adapter-spec.md` §0.1's D2 row and `okf-support.md`'s
  "v0.2 update (#730)" write-side paragraph **now describe the actual hybrid
  shape** (Decision 1) and note that stamping has no workflow exception
  (Decision 2) — both were still describing the retired fully-nested shape
  before this revision's docs-accuracy follow-up (#1 below).

## Owner decisions applied (this revision)

1. **On-disk frontmatter shape → hybrid.** See D2 above. Bare top-level
   `generated:`/`verified:` (OKF v0.2's own spelling); `sources` alone stays
   namespaced under `provenance:`. Round-trips through the `akm` adapter;
   `okf` adapter stays consumer-only and unaffected; `src/core/write-source.ts`
   untouched.
2. **Workflow allowlist now admits the stamped keys.** See D2 above. Closes
   the one undocumented coverage hole (workflow silently promoted unstamped);
   pinned by a new regression test; overstated inline comment corrected.

## Review follow-ups addressed (all small)

1. **Docs accuracy:** `okf-support.md`'s v0.2 write-side paragraph, the
   CHANGELOG entry, and the bundle-adapter spec §0.1 D2 row now describe the
   real hybrid shape and note stamping covers every AKM-native markdown type
   (no workflow exception). `IndexDocument.provenance`'s own doc comment
   (`types.ts`) got the same fix — it still said "nested under a
   `provenance:` key" for the write side.
2. **Corrected two counts below:** the per-file assertion-change split is
   **2/4/4/5** (not 2/3/4/6) across `proposals.test.ts` /
   `proposal-durable-recovery.test.ts` / `goldens-proposal-recovery.test.ts` /
   `goldens-proposal-txn.test.ts`; total 15 is unchanged and correct.
3. **CHANGELOG now notes** that promotion re-serializes frontmatter via
   `yaml.stringify`, so YAML comments in hand-written proposal frontmatter are
   not preserved (values and body bytes are).
4. **`capturedAtHead` guard added.** The stale-commit bug a human fixed by
   hand in the first revision (`fix(okf): repoint stale capturedAtHead...`)
   was never guarded against — and it recurred: *this very revision's own
   rebase onto `main`* rewrote commit hashes again and re-broke the same four
   OKF goldens' `capturedAtHead` pins. `scripts/lint-golden-captured-at-head.ts`
   (wired into `bun run lint`) now checks every golden's `capturedAtHead`
   exists as a real commit object and is reachable from at least one known
   branch (deliberately not "must be an ancestor of the current branch" —
   this repo rebases feature branches onto `main` routinely, which would
   flag nearly every legitimate historical pin). Repointed the four OKF
   goldens again.

   Caught its own bug before landing: every `.github/workflows/*.yml` job
   uses `actions/checkout@v5` with no `fetch-depth` override (the default is
   a `depth: 1` shallow clone), so the first version of this guard failed
   CI on ~11 goldens it never touched — a shallow clone genuinely can't tell
   "garbage-collected/orphaned" apart from "simply never fetched." Fixed by
   scoping the softening to exactly the ambiguous case: an ABSENT commit
   object is inconclusive (warns, doesn't fail) only when the clone is
   shallow; a commit that EXISTS but is unreachable from every branch — the
   actual bug class — still fails everywhere, shallow or not, since presence
   is conclusive regardless of depth. Verified by cloning at `--depth 1` and
   running the guard there.

   One pre-existing, unrelated violation the guard surfaced
   (`tests/fixtures/goldens/lint/all-types.json`, a chunk-3 golden with a
   `capturedAtHead` that doesn't exist in this clone at all) is grandfathered
   as a non-blocking warning rather than guessed at — out of scope for #730,
   flagged for its own surface owner instead of silently fixed or silently
   ignored.
5. **`okf-sample`'s README relocated.**
   `tests/fixtures/bundles/okf-sample/README.md` carried no frontmatter, so
   the `okf` adapter's directory-gate-free `recognize()` silently defaulted
   it to a `knowledge` concept inside a fixture meant to stay
   behavior-frozen. Moved to `tests/fixtures/bundles/okf-sample-README.md`
   (a sibling of the bundle root, not a member of it). No golden needed
   re-recording — grepped the whole tree first; nothing referenced "README"
   before this change.
6. **`human:<OS username>` left as-is**, per the owner's explicit call — it
   embeds the machine username into written assets and that's accepted for
   now. CHANGELOG now says so explicitly, so users know identifying data can
   land in content.

## Rebase onto `main`

`origin/main` advanced with PR #737 (#672, the salience/outcome repository
extraction) after this PR was first opened; both touch `CHANGELOG.md`.
Rebased onto `origin/main`, resolving the `CHANGELOG.md` conflict by keeping
both PRs' entries — verified both are present in the final file, each in its
own `### Added` list position. Re-ran `bun run lint` afterward: `main` now
carries a `state-table-sql` rule banning raw `asset_salience`/`asset_outcome`
SQL outside `src/storage/repositories/` — this PR's diff doesn't touch either
table, confirmed by `lint-repository-sql`'s own "OK" (see Test plan).

## Golden re-baselining

Decision 1 changes the written bytes, so
`tests/fixtures/goldens/journal/proposal-txn.json` (`re-baseline`-designated,
`DESIGNATIONS.json`) needed re-recording again: the same **2** `fileTree`
hashes the first revision's re-baseline already touched (new-asset and
overwrite-existing accept scenarios — the only two that hash real written
bytes) changed once more, reflecting the shape moving from fully-nested to
hybrid; every other recorded outcome (status/backup/event counts/journal
cleanup) stayed byte-for-byte identical. `DESIGNATIONS.json`'s notes field
documents this re-capture alongside the earlier one.

The **frozen** `tests/fixtures/goldens/journal/proposal-recovery.json` is
**untouched, sha256 pin intact** — its own comparator
(`contentMatchesAccepted`) only checks `description`/`when_to_use`/body
content, never full-file bytes or the provenance keys themselves, so it was
never sensitive to the shape change in either revision. Confirmed unchanged:
`git diff origin/main...HEAD -- tests/fixtures/goldens/journal/proposal-recovery.json`
is empty.

## Pre-existing test edits — justified, none loosened

Besides `proposals.test.ts`, three other files were touched only to
accommodate D2's new frontmatter block; here is why each pre-existing
assertion change is legitimate rather than a weakening:

`tests/integration/proposals.test.ts` (2 sites),
`tests/integration/proposal-durable-recovery.test.ts` (4 sites),
`tests/integration/goldens-proposal-recovery.test.ts` (4 sites), and
`tests/commands/proposal/goldens-proposal-txn.test.ts` (5 sites) — **15
sites total** — all changed `expect(fs.readFileSync(assetPath,
"utf8")).toBe()` (exact byte equality) into a structural comparison
(`expectAcceptedContent` / `contentMatchesAccepted` helpers, or a
snapshot-then-compare pattern) checking `type`/`description`/`when_to_use`/
body. This is a direct, unavoidable consequence of D2 intentionally adding a
new frontmatter block to accepted content — byte-identity with the pre-stamp
proposal was never going to survive a feature whose entire point is to
change what gets written. Each site is commented inline explaining exactly
this. I read every one of these diffs line-by-line rather than trusting the
commit message: none of them stopped checking anything about crash-recovery
correctness, backup integrity, revert-refusal, or accept/reject status — they
just stopped asserting full-file byte equality, which the feature
deliberately breaks. I did not find a case where this pattern was used to
paper over an actual bug.

## Test plan

All commands run in the worktree, on top of the final commit
(`732d69a6`) — real output, not asserted:

- `bun run lint` — **exit 0.** New `lint-golden-captured-at-head` check
  passes: 16 of 17 pins judged in this (also shallow) clone, all resolve to
  a real, reachable commit; the 1 unjudged pin is the pre-existing,
  unrelated `lint/all-types.json` violation, inconclusive-not-failing here
  because its commit object is simply absent (see follow-up #4). 1459 pre-existing
  `noNonNullAssertion`-style biome warnings repo-wide, confirmed unrelated to
  this change and not blocking (biome's exit code for warnings-only is 0);
  a direct `biome check` on every file this PR touches reports zero
  diagnostics.
- `bunx tsc --noEmit` — **exit 0, zero output.**
- `bun run test:unit` — **3020 pass / 0 fail across all 238/238 unit files**
  (4 process-shards, no coverage-hole).
- `bun run test:integration` — **5104 pass / 0 fail across all 380/380
  integration files** (4 process-shards, no coverage-hole).
- Targeted suites named for this review, all green:
  - OKF + conformance + fixture-consumer suites — `okf-adapter.test.ts`,
    `okf-conformance.test.ts`, `conformance.test.ts`, `registry.test.ts`,
    `reserved-filename-conformance.test.ts`, `akm-markdown.test.ts`,
    `indexer/installations.test.ts`,
    `integration/source-providers/provider-utils.test.ts` —
    **140 pass / 0 fail** (8 files).
  - `proposals.test.ts` + `proposal-durable-recovery.test.ts` —
    **92 pass / 0 fail** (2 files), including the new
    "workflow — the one closed-allowlist type — is stamped" test and the
    hybrid-shape-pinning assertions.
  - `goldens-proposal-txn.test.ts` + `goldens-proposal-recovery.test.ts` —
    **31 pass / 0 fail** (2 files, golden-consuming ripple).
  - `tests/workflow-markdown.test.ts` (workflow validator suite) —
    **14 pass / 0 fail** (1 file).
  - All 13 files together: **277 pass / 0 fail**, 963 `expect()` calls.

## Checklist

- [x] Tests pass (`bun test` — see full breakdown above: unit 3020/0,
      integration 5104/0)
- [x] Lint passes (`bun run lint` — exit 0)
- [x] Docs updated (runbook, bundle-adapter spec §0.1 mapping table,
      `okf-support.md` v0.2 note, CHANGELOG — all now describing the hybrid
      shape and the workflow-inclusive stamping coverage)

Fixes #730

---


itlackey (commented):
## Review — independent verification of every claim, plus two findings the PR body gets wrong

Reviewed from a clean clone at head `6fa806d1` against base `03b3a305` (verified identical to GitHub's live `main` via `git ls-remote`). Every check below was re-run or re-derived independently; nothing was taken from the PR body on trust. **No blocking defects found.** Two accuracy errors in the PR's self-description (both in the conservative direction), several non-blocking findings, and a recommendation at the end.

### Verified claims (all independently confirmed)

1. **Consumer-only scope holds — strongest possible confirmation.**
   - `git diff 03b3a305...6fa806d1 -- src/core/write-source.ts` is empty. `assertAkmAssetWrite` (write-source.ts:1318) still throws for any non-`akm` adapter.
   - Stamping sits at repository.ts:2242-2260, strictly after `assertAkmAssetWrite(mutationTarget.source)` at :2224, and only for `.md` paths. `stampProposalProvenance` has exactly one call site in the tree.
   - `okf-conformance.test.ts` has a single hunk `@@ -978,3 +978,107 @@` with zero deleted lines; I md5-compared lines 820-940 (the write-rejection block) between base and head — byte-identical.
   - The entire `repository.ts` diff deletes exactly 2 lines (lint + transaction now consume `stampedContent` instead of `repairedContent`); everything else is additive. Transaction/journal ordering is untouched.
2. **Golden re-baseline is legitimate.** `journal/proposal-txn.json` is `"designation": "re-baseline"` (`DESIGNATIONS.json:6`), not `frozen-migration-input`; exactly 2 fileTree hash lines changed; both capture sites pin `ctx: {now, actorId}` (`GOLDEN_CAPTURE_CTX`, goldens-proposal-txn.test.ts:655). The hashes are reproducible here — the suite does a strict `expectGolden` compare and passes. Bonus: `lint-goldens-presence` hash-verifies the 33 frozen goldens, and the frozen `journal/proposal-recovery.json` is byte-untouched with its sha256 pin matching.
3. **`capturedAtHead` fix verified.** All four `format-family-goldens/okf/*.json` uniformly record `15a1b3754e...`, which `git merge-base --is-ancestor` confirms is a HEAD ancestor. The stale `250f91bc...` does not even exist in a fresh clone (`git cat-file -t` → "Not a valid commit name"), proving the verifier's diagnosis (orphaned pre-amend object) and the necessity of the fix. Confirmed nothing in src/tests/scripts validates this field — recurrence is unguarded (recommendation below).
4. **`KC_REF` is real and current.** `git ls-remote https://github.com/GoogleCloudPlatform/knowledge-catalog.git HEAD refs/heads/main` → `3fcbb9f828c2f23d109c855ee403c3a4c81f3a96` for both. `okf/SPEC.md` fetched at that exact commit opens "**Version 0.2**"; I cross-checked details, all match the implementation: source-entry fields (`resource`/`id`/`title`/`author`/`usage_count`/`last_modified`, §5.1), `verified` single-mapping shorthand ("Consumers MUST treat a bare mapping as a one-element list", §5.2), `status: draft|stable|deprecated` (§5.4), `stale_after` (§5.5), and §13's "`timestamp` is superseded by `generated.at` … MAY fall back".
5. **v0.1 compatibility.** `okf-sample/` diff is exactly `A README.md`. One nuance the PR missed: the README has no frontmatter, and `recognize()` defaults frontmatter-less `.md` to `type: knowledge` — so the README *is* now an indexable concept of that fixture bundle. Empirically neutral: all five fixture-consuming suites pass, including the two the PR didn't list (`tests/indexer/installations.test.ts`, `tests/integration/source-providers/provider-utils.test.ts`). Consider frontmatter-marking or relocating the README if the fixture is meant to be behavior-frozen.
6. **Checks re-run from scratch on this machine:** `bun run lint` exit 0; `bunx tsc --noEmit` exit 0; OKF suites (the PR's six named files + the two fixture consumers above) **140 pass / 0 fail**; the four proposal/golden files **122 pass / 0 fail** (matches the PR's 91+31); full `bun run test:unit` **3020 pass / 0 fail (238/238 files)**; full `bun run test:integration` **5098 pass / 0 fail (380/380 files)** — both full suites exactly match the PR's claimed numbers, with no host-contention flakes in my run.
7. **CI:** all 5 check runs on `6fa806d1` completed **success** (`check`, `node-smoke (22)`, `node-smoke (24)`, `slow-gated-tests`, `smoke`); `mergeable_state` is now `clean` (the `unstable` at creation was checks-in-progress). Note #737 is in flight against the same base and also touches `CHANGELOG.md` — whichever merges second will need a trivial conflict resolution there.

### Finding 1 (PR body inaccuracy, benign direction): the closed-allowlist hole is exactly `{workflow}`, not `{workflow, command, agent, skill}`

The PR body and the inline comment (repository.ts:2232-2234) claim command/agent/skill "share the same closed-frontmatter posture" as workflow. **That is wrong for this tree.** The only closed frontmatter allowlist in `src/` is the workflow one (`src/workflows/validator.ts:81-88`: `type, title, description, tags, params, name, timestamp, updated, when_to_use, xrefs`), and the canonical proposal-validator dispatch covers only `lesson`/`task`/`workflow` (`proposal-validators.ts:67-100`) — command/agent/skill get only the generic + quality validators, which tolerate `provenance:`. I confirmed empirically with a scratch matrix test promoting one proposal of every type:

| Outcome | Types |
|---|---|
| **Stamped** | lesson, knowledge, memory, fact, session, instruction, **command, agent, skill** (9) |
| **Silently unstamped** (fallback, one `warn`) | **workflow** — and only workflow |
| Never attempted (non-markdown, no frontmatter) | task (.yml), script, env, secret |

So actual feature coverage is *better* than documented, and the safety property holds (the workflow fallback promotes unstamped content rather than failing; `promotionLintBlockers` only blocks on `unquoted-colon`/`missing-ref`/`stale-path`, so no late rejection path exists either). Remaining issues, all non-blocking:
- The inline comment overstates the hole — fix it to name workflow alone (repository.ts:2232-2234).
- **No user-facing doc mentions the workflow exception at all**: `okf-support.md`'s v0.2 write-side paragraph, the CHANGELOG entry, and the §0.1 D2 row all say provenance is stamped on AKM-native (markdown) writes, unqualified. One sentence in each would close the gap.
- No test pins the workflow-fallback outcome (unstamped + warning). Pre-existing tests only prove workflow promotion doesn't *fail*. Worth a small test so the fallback isn't accidentally lost or silently widened.
- The fallback warning is invisible in `--quiet` runs (log-file only) — acceptable, but worth knowing for the drain/triage path.

### Finding 2 (PR body arithmetic): 15 assertion sites confirmed, but the per-file split is 2/4/4/5, not 2/3/4/6

I read all of them line-by-line. `proposals.test.ts` 2, `proposal-durable-recovery.test.ts` **4** (:113, :147, :204, :231), `goldens-proposal-recovery.test.ts` 4, `goldens-proposal-txn.test.ts` **5** structural conversions (+2 separate `ctx`-pinning edits, which is presumably how the PR counted 6). Total 15 — the headline number is right; the attribution is off by one in each of two files.

**Independent judgment on whether any site would now miss corruption — the question that matters:**
- Every helper (`expectAcceptedContent`, `contentMatchesAccepted`) retains **full body byte-equality** (`a.content === p.content`), so a silently truncated or altered promoted body is still caught at every one of the 15 sites. Frontmatter is compared field-wise (`type`/`description`/`when_to_use` — the complete set these fixtures carry).
- Backup-swap: the neighboring, unchanged assertions still pin it exactly (`proposal.backupContent).toBe(seeded.original)` at durable-recovery:117-119 and the golden's `backupContentMatchesOriginal`, still strict `===`). Status, event-count, journal-cleanup, and revert-refusal assertions are all unchanged.
- The two `proposals.test.ts` sites keep full byte-equality for the property under test (refused revert leaves the file untouched, `toBe(afterBAccept)` against a post-accept snapshot); the snapshot itself is sanity-checked with `toContain` markers that cover the one-line fixture bodies.
- **The real residual gap** is narrow and compensated: the helpers tolerate arbitrary *additional* frontmatter keys beyond `provenance` (by design), so hypothetical garbage keys added at accept time would pass these 15 sites. But `goldens-proposal-txn.test.ts` still captures `fileTree` **SHA-256 of the accepted asset's raw bytes** into a strict golden — byte-exact accept output remains pinned there, so this gap has a byte-exact backstop.
- One nuance worth stating out loud: the **frozen** `journal/proposal-recovery.json` survives byte-for-byte only because its consumer redefined the recorded booleans' *predicate* (byte-equality → structural). The fixture and its sha256 pin are untouched and the booleans still verify "content survived recovery" (including body byte-equality), so I judge this the correct move — re-recording a frozen golden would have been far worse — but it is a semantic reinterpretation of a frozen oracle and deserves this explicit acknowledgment.

Verdict: none of the 15 changes papers over a bug; none loses truncation/backup/status coverage. The conversion is honest and the byte-exact backstop survives.

### The two flagged design decisions — my independent judgment for the owner

- **D1.3 naming (`IndexDocument.provenance` + flat `lifecycleStatus`/`staleAfter`/`okfVersion`):** exactly the plan's own recommendation; I verified the three adjacent pre-existing fields (`sources?: string[]`, `generation?: number`, `quality: "generated"`) are untouched in `types.ts`. No concerns — endorse.
- **On-disk shape (nested `provenance:` vs the plan's literal bare `generated:`/`verified:`):** the collision argument is real but applies only to `sources:` — `applyWikiFrontmatter` (metadata.ts:468+) consumes bare `sources:` on all akm markdown and silently drops non-strings. I verified bare `generated:`/`verified:` have **no** existing consumer in the read path, so a hybrid (spec-shaped `generated:`/`verified:` + namespaced sources) was available. The tradeoff the owner is actually deciding: all-nested keeps one coherent key and mirrors `IndexDocument.provenance`, but means a third-party OKF v0.2 consumer reading an AKM stash as an OKF bundle (the "AKM markdown is an OKF-compatible superset" posture in `okf-support.md`) sees no trust metadata — `provenance:` is just an unknown extra key to it. The implemented choice is coherent, collision-safe, and migratable later; the hybrid would have bought spec-visible trust metadata at the cost of a split shape. Either is defensible — this is genuinely the owner's call, and the frontmatter shape is easy to change before assets accumulate but annoying after.

### Other non-blocking observations

- `stampProposalProvenance` re-serializes the **entire** frontmatter via `yaml.stringify` (asset-serialize.ts:48), so promotion of stamped types now normalizes YAML formatting and would drop YAML comments in proposal frontmatter — values preserved, body bytes preserved (proven by the helpers), but the byte effect is "frontmatter rewritten + block appended", slightly broader than the PR's "adds a frontmatter block". Mostly theoretical for machine-generated proposal content; worth one line in the CHANGELOG if anyone hand-writes proposals.
- `human:<OS username>` fallback (repository.ts:1961-1969) embeds the machine username into written assets for human-sourced/human-accepted promotions. Aligned with the spec's `human:<id>` convention and seam-overridable, but it is machine-identifying data in content — flagging for owner awareness.
- Malformed values of the six consumed v0.2 keys (e.g. `status: archived`) are now dropped entirely rather than surviving in `documentJson` extras. Consistent with the v0.1 treatment of consumed keys; noting for completeness.
- `capturedAtHead` recurrence guard: recommend a tiny lint (existence-check `git cat-file -e` per committed golden's `capturedAtHead`, ancestry where cheap) — the exact failure the verifier caught by hand would then be mechanical.

### Recommendation

**Merge as-is once the owner confirms the two consulted questions** — (1) the D1.3/on-disk `provenance:` nesting as implemented, with the OKF-consumer-visibility tradeoff stated above in view, and (2) that closed-allowlist types silently falling back to unstamped (with the correction that this is *only* workflow) is the desired behavior. No pre-merge code change is required; if the owner's answers match the implementation, land it, then fix in a small follow-up: the repository.ts:2232 comment, one-sentence workflow-exception notes in CHANGELOG/`okf-support.md`/spec §0.1, a workflow-fallback pin test, and optionally the `capturedAtHead` lint. If the owner instead wants the bare/hybrid on-disk shape or strict-fail behavior for allowlist types, that is a design change to this PR, not a follow-up.

Mind the #737 CHANGELOG merge-order conflict; whichever lands second needs a trivial rebase.

---

itlackey:
Both consulted decisions are now applied, plus the review's small follow-ups. Rebased onto `main` (which carries #737) and pushed as `0e13e289`.

## 1. On-disk shape → hybrid

`generated:` and `verified:` are now stamped **bare at the top level**, exactly as OKF v0.2 spells them; only `sources` stays namespaced under `provenance:`.

The review's own finding decided this: bare `generated:`/`verified:` have **no pre-existing consumer** in the read path, so the fully-nested shape bought nothing while leaving a third-party OKF v0.2 reader with a single unrecognized `provenance:` key instead of conformant trust metadata — which `okf-support.md`'s OKF-superset positioning, decided in *this* release, promises. Only `sources` genuinely collides (with `applyWikiFrontmatter`'s citation-**string** convention), so only it is namespaced. Stamped frontmatter is durable state in user stashes, so this is settled before assets accumulate rather than via a later migration.

Read side (`applyProvenanceFrontmatter`) reads the bare families and still accepts the older nested spelling as a fallback, so the round-trip into `IndexDocument.provenance` is unchanged. Two behavior improvements fell out: `verified` now **accumulates** across re-promotions rather than the newest entry replacing the record, and an otherwise-empty `provenance:` block is dropped instead of left as a dead key.

`IndexDocument.provenance` / `lifecycleStatus` / `staleAfter` / `okfVersion` keep their TypeScript shape — only the on-disk spelling changed.

## 2. Workflow is stamped, not silently skipped

Thank you for the correction — it was right, and it made this fix one line instead of four. `workflow` is the **only** type with a closed frontmatter allowlist, so it was the sole type promoting unstamped through the re-validation fallback. Its allowlist now admits the three keys. The fallback is retained purely as a backstop for any future closed-allowlist type, and the overstated comment at `repository.ts` naming command/agent/skill is corrected.

**The new test provably depends on the fix:** with the three keys removed from the allowlist it fails (`generatedBy: undefined`); restored, it passes. Also added a test pinning the on-disk *spelling* — bare keys present, `provenance:` absent — so a regression to the nested shape can't pass silently.

## CI failure on the rebase commit was the known flake

`check` failed on `13eccc11` with one test: `already-current proposal ref repair > resumes after crashing immediately after current proposal repair commit` (`three-db-cutover.test.ts`) — the same test #737's verifier identified as a host-contention flake. Since it is a *proposal-recovery* test and this PR changes what promotion writes, I did not assume: it passed **3/3 in isolation**, and the full suite below is green including it.

## Verification (real output, this branch at `0e13e289`)

- `bun run lint` → exit 0 · `bunx tsc --noEmit` → exit 0
- `bun run test:unit` → **3020 pass / 0 fail** (238/238 files)
- `bun run test:integration` → **5104 pass / 0 fail** (380/380 files)
- Golden: re-recorded the `re-baseline`-designated `journal/proposal-txn.json` (exactly 2 `fileTree` hashes — written bytes changed by design) via `AKM_UPDATE_GOLDENS=1`, verified a clean re-run, and updated its `DESIGNATIONS.json` notes to describe the hybrid shape. The **frozen** `journal/proposal-recovery.json` is byte-untouched.

## Also folded in

Docs (`okf-support.md`, CHANGELOG) now describe the real shape and state that every AKM-native markdown type is stamped, workflow included. The CHANGELOG also records the two consequences the review surfaced: promotion re-serializes frontmatter so YAML **comments** in hand-written proposals are not preserved (values and body bytes are), and the `human:<OS username>` fallback puts that username into written content.

## Not done

The `capturedAtHead` lint guard the review recommended is **not** in this push. It is a genuinely good idea — the stale-commit bug was caught only by hand — but it guards a free-form audit field rather than this PR's behavior, so it belongs in a follow-up rather than growing this diff further. Same for relocating the `okf-sample/README.md` that now indexes as a `type: knowledge` concept: harmless today (all fixture suites pass), worth tidying separately.

---

itlackey:
Applied both owner decisions and all six review follow-ups, on top of a rebase onto `main` (PR #737 merged in the meantime).

**Decision 1 — hybrid on-disk shape.** `generated:`/`verified:` now stamp bare at the top level (OKF v0.2's own spelling); only `sources` stays namespaced under `provenance:` (it's the one field that actually collides with the wiki citation-string convention). Round-trips: `applyProvenanceFrontmatter` reads the new bare keys (plus the old nested spelling as a fallback) into the same `IndexDocument.provenance` shape; the `okf` adapter's own read side is unaffected and lands in the same shape; consumer-only invariant re-verified (`write-source.ts` untouched, stamping still gated behind `assertAkmAssetWrite`, `.md`-only).

**Decision 2 — workflow is no longer the coverage hole.** The review's correction was right: `workflow` is the *only* type with a closed frontmatter allowlist, so it was the one type silently promoted unstamped. `ALLOWED_FRONTMATTER_KEYS` now admits the stamped keys, and a new test (`"workflow — the one closed-allowlist type — is stamped, not silently skipped"`) pins it — verified it actually depends on the fix by removing the keys and watching it fail first.

**Follow-ups:** docs (okf-support.md, bundle-adapter spec §0.1, and the `types.ts` doc comment) now describe the real hybrid shape; the PR body's assertion-count split is corrected to 2/4/4/5 (total 15 unchanged); CHANGELOG notes both the YAML-comment-loss consequence of re-serializing frontmatter and the `human:<OS username>` fallback (left as-is per the owner's call); added a `capturedAtHead` integrity guard (`scripts/lint-golden-captured-at-head.ts`, wired into `bun run lint`) after finding the OKF goldens' pins had gone stale *again* from this session's own rebase — it checks existence + branch-reachability rather than ancestor-of-HEAD (which routine rebasing would make everyone fail), and was itself corrected to not fail CI's shallow `fetch-depth: 1` checkout while still catching the real bug class (object present but unreachable) regardless of clone depth; and the `okf-sample` fixture README (previously frontmatter-less, so the directory-gate-free `okf` adapter silently indexed it as a stray `knowledge` concept) is relocated beside the bundle instead of inside it.

`proposal-txn.json`'s two `fileTree` hashes were re-baselined again for the shape change (it's `re-baseline`-designated); the frozen `proposal-recovery.json` is untouched — its comparator never looked at exact bytes or the provenance keys, so it was never sensitive to this.

Full verification (real output, not asserted): `bun run lint` exit 0, `bunx tsc --noEmit` clean, unit 3020/0 (238/238 files), integration 5104/0 (380/380 files), plus all 13 suites named in review (OKF/conformance/fixture-consumer, proposal core, proposal goldens, workflow validator) — 277/0. Full breakdown in the updated PR description.

---

itlackey:
## Re-review of the delta (6fa806d1 → 732d69a6 / 4a9a9c59) — merged

Reviewer of record, second round. Both owner decisions from the first round were applied; I re-verified the delta in a scratch clone at `732d69a6` (and confirmed `4a9a9c59` is an empty retrigger commit — tree hash `c0f92c83…` identical to `732d69a6`, so every finding below carries over to the merged head).

### 1. Hybrid shape round-trips
`stampProposalProvenance` (`src/commands/proposal/repository.ts`) writes bare top-level `generated: {by, at}` and `verified: [{by, at}]` with only `sources` namespaced under `provenance:`; the older nested keys are migrated/deleted and an empty `provenance:` block is dropped entirely. `applyProvenanceFrontmatter` (`src/indexer/passes/metadata.ts`) reads the bare families first with the fully-nested spelling as fallback, folding both onto the unchanged `IndexDocument.provenance` TypeScript shape (the types.ts delta in `1d3d6ad` is comment-only — I filtered the diff for non-comment lines: none). `verified` accumulates: prior entries are absorbed from either spelling and appended to, so re-promotions extend the record instead of replacing it. `generated` is replaced each promotion, which is correct — it describes the current content. One cosmetic edge noted, not blocking: a proposal whose own hand-written content uses v0.2's single-mapping `verified:` shorthand is not absorbed into the list on first promotion (only arrays are); everything the stamper itself writes is a list, so accumulation across re-promotions is unaffected.

### 2. Consumer-only scope — no regression
- Stamping begins at `repository.ts:2266`, strictly after `assertAkmAssetWrite(mutationTarget.source)` at `:2246`.
- `git diff origin/main...HEAD -- src/core/write-source.ts` — empty.
- `okf-conformance.test.ts` diff vs main is a single EOF-append hunk (`@@ -978,3 +978,107 @@`, the new v0.2 fixture block); the write-rejection block is byte-identical to main.

### 3. Workflow claim — bite experiment performed
The new test (`proposals.test.ts:1544`, "workflow — the one closed-allowlist type — is stamped, not silently skipped") genuinely depends on the allowlist fix. I removed `generated`/`verified`/`provenance` from `ALLOWED_FRONTMATTER_KEYS` (`src/workflows/validator.ts`) and the test failed with `expect(provenance.generatedBy).toBe(...)` receiving `undefined` — the exact claimed failure mode (re-validation fallback promoted unstamped). Restored the keys; test green again, working tree byte-clean. The on-disk spelling is separately pinned (`proposals.test.ts:1534-1541`): bare `generated`/`verified` asserted at top level and `fm.provenance` asserted `undefined` when it would be empty, so a silent regression to the nested shape cannot pass.

### 4. capturedAtHead guard — teeth verified, not a mask
All three behavior quadrants tested directly:
- **Absent object + depth-1 shallow clone:** fresh `git clone --depth 1`, ran `bun scripts/lint-golden-captured-at-head.ts` — exit 0, reporting "0 of 17 capturedAtHead pin(s) checked … 17 unjudged (shallow clone)". A shallow run cannot read as full coverage.
- **Present-but-unreachable + shallow clone (the #730 bug class):** fabricated a dangling commit via `git commit-tree` in the same shallow clone (object type `commit`, zero containing branches) and pointed a fixture golden at it — guard failed with exit 1 and the "exists locally but is unreachable from every branch" message. The conclusive check keeps its teeth even at depth 1; the guard is not neutered.
- **Absent object + full clone:** fabricated a nonexistent-SHA pin in the full clone — guard failed with exit 1.

The softening in `7a70f7a1` is scoped to exactly the inconclusive case. One incidental finding: in a genuinely complete clone the grandfathered `lint/all-types.json` pin actually resolves and is reachable (guard reports 17 of 17, zero warnings) — the KNOWN_STALE entry reflects the introducing author's incomplete clone rather than a real repo violation. Warn-only either way; its surface owner can simply drop the entry.

### 5. Goldens
- `journal/proposal-txn.json`: diff vs main is exactly the 2 `fileTree` hashes claimed (new-asset + overwrite accept scenarios), nothing else; `DESIGNATIONS.json` notes document the re-capture, the hybrid shape, the `ctx: {now, actorId}` reproducibility seams, and the honest caveat that a version bump will shift these hashes again.
- Frozen `journal/proposal-recovery.json`: byte-untouched vs main; raw-bytes sha256 equals the DESIGNATIONS pin (`30cd1dce…`), and `lint-goldens-presence` independently hash-verifies all 33 frozen goldens.
- The four OKF format-family goldens all pin `23ccdbde` (rebased D2 GREEN) — a real commit object, reachable from the PR branch, and an ancestor of the merged head, so it stays reachable from `main` after merge.

### 6. Checks run (scratch clone at 732d69a6)
- `bun run lint` — exit 0 (all 12 lints incl. the new guard at 17/17; the 1459 biome warnings are the documented pre-existing set).
- `bunx tsc --noEmit` — exit 0, zero output.
- Targeted suites: OKF/conformance/adapter files 111 pass / 0 fail; akm-markdown + akm-adapter + workflow-markdown 44/0; `proposals` + `proposal-durable-recovery` 92/0; both golden consumers 31/0. All match the PR body's claimed counts.

### CI flake determination
The single `check` failure on `732d69a6` (`proposal-storage-sqlite.test.ts` > "concurrent reject + gate-decision mutation cannot revive a pending row") was verified as a pre-existing load flake, not a regression: the file is not in this PR's diff, the test is an explicit two-worker race whose assertions branch on either interleaving (`if (gateOutcome.updated) … else …`), and it passed 5/5 consecutive isolated runs here. The empty retrigger commit `4a9a9c59` came back fully green — all 5 check runs (`check`, `slow-gated-tests`, `node-smoke (22)`, `node-smoke (24)`, `smoke`) succeeded, so the flake did not recur.

Merged with merge commit `9f3c8f13e13369b3c47a7f32a08aaa2354ab09b7`, deliberately ahead of #739 so the shared `CHANGELOG.md` rebase falls to the smaller change.

---

Issue #585, opened 2026-06-10
fix(improve): reuse eventsCtx.db in post-loop purge instead of opening second state.db connection


itlackey:
## Problem

`improve.ts` opens `eventsDb` (via `eventsCtx`) at ~line 1179 as a long-lived write connection held for the entire improve run. The post-loop purge block at ~line 3263 opens a *second* `stateDb = openStateDatabase()` for `purgeOldEvents`/`purgeOldImproveRuns` while `eventsDb` is still live.

In WAL mode, two simultaneous write handles to the same SQLite file plus external writer contention (cron tasks at `:00`/`:15` etc.) causes `SQLiteError: database is locked`.

**Observed:** `[akm] task history DB write failed: SQLiteError: database is locked` repeated 6+ times per `discord-wiki-articles-ingest` run; post-loop purge failures in improve-quick runs.

## Fix

Reuse `eventsCtx.db` in the post-loop purge:

```ts
// improve.ts ~3263
const stateDb = eventsCtx.db;
if (stateDb) {
  purgeOldEvents(stateDb, retentionDays);
  purgeOldImproveRuns(stateDb, retentionDays);
}
```

Eliminates the second writer entirely and closes the race window.

## Related

Part of a trio with #584 (index.db double-open) and the busy_timeout increase. Together they eliminate all observed `database is locked` failures.

itlackey:
Verified implemented: post-loop purge reuses eventsCtx.db (ownsStateDb guard) instead of opening a second state.db. src/commands/improve/improve.ts:3866-3921. Commits 1f8c5cf2 / bbe12b6d. (Independent re-verification.)

PR #739, opened 2026-07-30
feat(improve): orphan-GC pass for unresolvable state rows (#733)

### LINKED ISSUE #733: improve: orphan-GC pass for state rows whose refs no longer resolve

itlackey:
## Summary

Implements #733 — an orphan-GC maintenance pass for `asset_salience` /
`asset_outcome` state rows whose refs no longer resolve against
`entries.item_ref`, per Workstream C of
`docs/architecture/specs/0.9.0-close-out-plan.md`.

**The design is deliberately lean — exactly four things**, matching the
owner's final scoping (issue #733, comment
[#5127316435](https://github.com/itlackey/akm/issues/733#issuecomment-5127316435)):

1. **One maintenance pass** — `runOrphanStateGcPass` in
   `src/commands/improve/loop-stages.ts`, registered in the maintenance
   sequence next to `runOrphanProposalPurgePass`. Per run: collects distinct
   refs from both state tables (via the repositories), resolves each against
   `entries.item_ref` (reusing the existing `getEntryByRef` single-ref probe,
   plus a bare-conceptId fallback via `bareImproveRef` for a stale
   bundle-qualified spelling), stamps `missing_since` on newly-unresolved
   refs, clears it on refs that resolve again, and — only when
   `improve.stateGc.collect` is `true` — deletes rows past the grace window,
   one `DELETE` per table.
2. **One additive migration** — `021-asset-state-missing-since`
   (`src/core/state/migrations.ts`): two `ALTER TABLE … ADD COLUMN
   missing_since INTEGER DEFAULT NULL` statements, nothing else.
3. **One event type** — `asset_state_gc` (`src/core/events.ts`), sentinel ref
   `asset_state/_gc`, metadata `{pending, collected, byTable}`. Emitted only
   when there's something to report; a clean run emits nothing.
4. **One config gate** — `improve.stateGc.collect: boolean`
   (`src/core/config/schema/improve.ts`), **default `false`**. The pass
   always runs and always reports counts; only physical deletion is gated,
   so live data can prove the report clean before the default ever flips.

**Explicitly NOT built** (all raised and rejected during the issue's own
scoping discussion): no quarantine archive, no circuit breaker, no
health-advisory plumbing, no new tables, no grace-period config key (the
7-day grace window is the named constant `STATE_GC_GRACE_MS`, mirroring
`TXN_SWEEP_GRACE_MS`, not configurable). `usage_events` is out of scope —
attached rows already cascade via `deleteUsageEventsByEntryIds` and the
90-day `purgeOldUsageEvents` bounds detached remnants; it is deliberately not
added to the `state-table-sql` lint rule either.

"Absent ≠ deleted" is inherited from the indexer, not re-implemented: an
incomplete or failed source scan preserves that source's last-known-good
`entries` rows (`indexer.ts` ~1195-1199), and mass-wipe is already gated
upstream. So "ref not present in `entries.item_ref`" is the entire
authoritative-deletion predicate — no separate scan-status tracking was
added.

Fixes #733

## Commit order (strict test-first / RED → GREEN)

1. `a1909faa` **RED** — the full test suite
   (`tests/integration/commands/improve/state-gc.test.ts`) plus the minimal
   non-algorithmic scaffolding it needs to compile and run: migration 021,
   the four CRUD helpers per repository (list/stamp/clear/delete/count on
   `missing_since`), the config schema, and a stub `runOrphanStateGcPass`
   that always returns zero and touches nothing. 10 of 13 tests failed on
   real behavioral assertions against the stub (not import errors); the
   other 3 asserted "should not happen" properties a no-op trivially
   satisfies.
2. `f20a598d` **GREEN** — the real `runOrphanStateGcPass` implementation,
   the `asset_state_gc` event, and registration in the maintenance sequence.
   All 13 tests pass.
3. `8305ee04`, `2061bf59` — fixes to three pre-existing migration-fixture
   test files (`migration-lifecycle-regression.test.ts`,
   `migration-backup.test.ts`, `migration-apply-crash.test.ts`) whose
   hand-rolled ledgers claimed migrations 009/011/015 (asset_salience) were
   already applied without ever materializing the physical table — harmless
   until migration 021 fell into the replayed range and needed the table to
   `ALTER`. Same fix shape each time: add the table's cumulative
   pre-cutover-checkpoint DDL, mirroring the file's existing
   `ASSET_OUTCOME_010_DDL` precedent. Also updated two ledger-tip assertions
   in `three-db-cutover.test.ts` and
   `sqlite-migrations.characterization.test.ts` (+ its snapshot) for the new
   migration count, and two relative-position `.at(-2)`→`.at(-3)`
   fixture-ceiling cross-checks in `migration-fixtures.test.ts`.

## How the legacy-spelling trap was handled

The naive `asset_ref NOT IN (SELECT item_ref FROM entries)` sweep the issue
warned against would delete live rows whenever a state row is keyed under a
different spelling than the live `entries.item_ref` (e.g. a bare conceptId
written before the asset was bundle-resolved). Instead, resolution goes
through the same per-ref `getEntryByRef` probe the rest of improve already
uses for reads — its bundle-parse + bare-conceptId-suffix-match logic
already handles that case — plus a `bareImproveRef` fallback (the same
primitive `preparation.ts`'s `normalizeStoredKey` map is built from) for a
stale bundle-qualified spelling. Test: `"a live asset whose state row
carries a legacy (bare) ref spelling is not a candidate"` in
`state-gc.test.ts` constructs exactly this scenario (state row keyed bare,
`entries.item_ref` bundle-qualified) and asserts it is never stamped.

## Test plan

New suite (`tests/integration/commands/improve/state-gc.test.ts`, 13 tests,
calling `runOrphanStateGcPass` directly with injected db handles — no LLM, no
full `akmImprove` run):

- Orphaned ref stamped (not deleted) on first sighting.
- Stamped orphan deleted only after the grace window elapses **and**
  `collect: true` (old `missing_since` seeded directly — no sleeping, per
  `scripts/lint-tests-isolation.ts`).
- Stamped orphan inside the grace window left alone even with `collect: true`.
- Default config (`collect` absent): nothing ever deleted; counts still
  reported.
- Legacy-spelling trap (above).
- A ref that resolves again after being stamped has `missing_since` cleared.
- A source whose scan did not complete contributes no candidates (modeled by
  an `entries` row present regardless of on-disk reality — the indexer's own
  suite covers *why* the row survives an incomplete scan).
- `asset_outcome` orphans tracked independently from `asset_salience`.
- Runs correctly through a borrowed (already-open) state.db connection
  (#585 discipline) without closing a handle it doesn't own.
- No index.db handle → warning, not a throw.
- `asset_state_gc` event: silent on a clean run; correct
  `{pending, collected, byTable}` metadata when something's pending; correct
  `collected` count once the grace window elapses.

Verification (all commands run for real in this branch's worktree):

- `bun run lint` — **exit 0** (includes `state-table-sql` and the
  `schemas/akm-config.json` `--check`).
- `bunx tsc --noEmit` — **exit 0**, no output.
- `bun run test:unit` — **2986 pass / 0 fail** across 238 files.
- New suite alone — **13 pass / 0 fail**.
- `tests/integration/commands/improve/*.test.ts` (24 files) — **265 pass / 0
  fail**.
- Migration suites (`migration-lifecycle-regression`, `migration-backup`,
  `migration-apply-crash`, `three-db-cutover`, `migration-fixtures`,
  `sqlite-migrations.characterization`) — **all green** in combined and
  isolated runs.
- Full `bun run test:integration` (381 files, 4 shards): run after the
  fixture fixes above. One `vault-dangerous-key-install-gate.test.ts`
  failure seen in an earlier sharded run (before these fixes were complete)
  was confirmed a pre-existing, unrelated flake — clean 7/7 rerun in
  isolation.

## Checklist

- [x] Tests pass (`bun test`)
- [x] Docs updated — CHANGELOG.md `Added` entry

---


itlackey:
Reviewed and merged as the final Workstream C / 0.9.0 gate. What was verified, in order of risk:

## 1. The legacy-spelling trap — adversarial analysis (the main event)

Static analysis first: `isStateRefLive` = `getEntryByRef(storedRef)` (exact `item_ref` match when bundle-qualified, byte-literal `substr`-suffix `//conceptId` match when bare, `.md`-toggle variants, NFC-normalizing parse) with a `bareImproveRef` fallback for stale bundle prefixes. Then I wrote and ran a throwaway adversarial suite against the real pass — every case with `collect: true` **and a pre-aged (grace+1min) `missing_since` stamp seeded**, so any wrong "unresolved" verdict would have physically deleted the row:

| Case | Setup | Outcome |
|---|---|---|
| Stale bundle prefix | state `oldbundle//memories/foo`, live entry `newbundle//memories/foo` | **kept**, stamp cleared (bare fallback resolves) |
| Shared conceptId, one bundle's asset deleted | state `b1//shared/foo` (b1's asset gone), live `b2//shared/foo` | **kept**, un-stamped (conservative false-negative, by design) |
| Path-style non-slug origin | state `~/code/foo//bar`, aged stamp, `collect: true` | pass **fails closed**: warning, zero deletions — even a genuine aged orphan in the same run survives |
| `.md` toggle | state `fixture//memories/note.md`, entry `…//note` | **kept** (`withMdVariants`) |
| NFD vs NFC unicode | state NFD spelling, entry NFC | **kept** (parse-level NFC normalization — a raw SQL `NOT IN` would have deleted this) |
| Suffix under-match control | state bare `foo`, only `fixture//sub/foo` live | correctly **stamped** (different conceptId), grace still protects it |
| Deletion control | genuine aged orphan + live sibling | orphan collected, sibling untouched |

**7/7 — no constructed spelling produced a false positive.** Every ambiguity resolves toward "keep"; the only failure mode of an exotic ref is fail-closed (warn + no-op for the whole sweep).

On reachability of the path-style case: the three-db cutover's `classifyCutoverRef` re-keys mapped refs, **quarantines** parseable legacy refs (incl. all `origin//type:name` path spellings — `classifyRefGrammar` calls any non-slug prefix "legacy"), and fails closed on unparseable ones — so the 2026-07-28 field-failure spellings cannot survive into a post-cutover live table, and current writers only produce `itemRef ?? conceptId` keys. The parse-throw path is manual-SQL-only territory, and it deletes nothing.

Residual I could not fully rule out (documented, not blocking): an `entries` row whose `item_ref` is still NULL (a write-back straggler) is invisible to both probe arms, so its state row would be stamped. It would additionally take `collect: true` (default **off** all of 0.9.0) plus the straggler surviving un-healed for 7 days of improve cycles (each of which indexes first) to become a deletion — and `indexedProvenanceForFile` throws on NULL `item_ref` today, so current writers cannot mint such rows.

## 2-7, briefly

- **Scope**: exactly the four contracted pieces (pass / migration 021 / `asset_state_gc` / `improve.stateGc.collect`). Grep-swept the diff for a fifth concept: quarantine, circuit breaker, advisory, new tables, grace config key, `usage_events` — only prose mentions saying "not built". No creep.
- **Absent ≠ deleted**: inherited, not reimplemented — `indexer.ts` `if (!walkComplete) … preserving its last-known-good rows` confirmed in the current tree; the pass adds zero scan-status tracking. Coverage composes: `zero-document-preflight.test.ts` proves preservation against a real index run with an unreadable root; `state-gc.test.ts` proves presence ⇒ zero candidates.
- **Grace/collect gate**: deletion requires `collect === true` **and** `missing_since IS NOT NULL AND missing_since < now - STATE_GC_GRACE_MS` (repository SQL cannot touch NULL-stamp rows). Grace tests seed aged stamps directly; `lint-tests-isolation` passes.
- **Migration 021**: two additive `ADD COLUMN … DEFAULT NULL`, nothing else. Fresh DBs replay 001→021 (009/010 create the tables; nothing before 021 ever drops them — 018 only drops *other* tables + an outcome column), so "no such table" is impossible on any legitimate replay path; the characterization snapshot confirms the fresh-path end state.
- **Repository boundary**: all raw state-table SQL sits in the two repositories; `state-table-sql` verified genuinely active by canary (injected a violation into `src/commands/improve/shared.ts` → rule fired with file/line; reverted).
- **Wiring**: registered in `runMaintenancePassesUnderLease` between the orphan-proposal purge and proposal expiration, on the shared `dbCell.current` index handle, state.db via `withStateDb(…, { path: eventsCtx?.dbPath, borrowed: eventsCtx?.db })` — the #585 discipline; the borrowed-handle test asserts the pass doesn't close what it doesn't own.

## Pre-existing fixture edits (commits 8305ee04, 2061bf59) — judgment: genuine realism fixes, nothing relaxed

The three fixtures' hand-rolled ledgers claimed 009/011/015 applied without materializing `asset_salience` — a state no real DB can reach (a ledger only says 009 ran if 009 ran, and nothing drops the table afterward). Each fix adds the cumulative 009+011+015 DDL (verified column-for-column against the real migrations, CREATE-then-ALTER order included), mirroring each file's own pre-existing `ASSET_OUTCOME_010_DDL` precedent for the identical migration-018 problem. The two hand-rolled narrow `CREATE TABLE asset_salience` snippets in `migration-lifecycle-regression.test.ts` became plain INSERTs — their `:bad`-ref cutover-failure purpose is intact. Zero assertions weakened; `migration-apply-crash` and `migration-backup` diffs are insertion-only. `three-db-cutover.test.ts` is now **stronger**: the old head-pin (`.at(-1) === "020"`) became `toContain("020-three-db-cutover")` (order-independent presence — a future migration can no longer mask the cutover's absence) *plus* a head-pin at 021. `migration-fixtures.test.ts`'s `.at(-2)`→`.at(-3)` are relative cross-checks whose absolute anchors (`"019-proposal-fingerprints"`, `PRE_CUTOVER_STATE_CEILING`) are unchanged.

## Runs (all local, on this branch; re-verified on the post-#738 merged tree)

- `bun run lint` — exit 0 (incl. `state-table-sql`, `lint-tests-isolation`, schema `--check`, and post-merge #738's `lint-golden-captured-at-head` 17/17)
- `bunx tsc --noEmit` — exit 0
- `bun run test:unit` — 2986/0 (238/238 files)
- `state-gc.test.ts` — 13/0; adversarial probe suite — 7/0 (deleted after the run, not part of the PR)
- `tests/integration/commands/improve/` — 265/0 (24 files)
- All six migration suites together — 184/0
- Post-#738-merge combined tree: lint 0, tsc 0, `state-gc` + `proposals` 88/0

## CI

Branch updated onto post-#738 `main` via the API (no hand-edits; the CHANGELOG union auto-merged, both entries verified present). The update-merge head `f16f1f45` failed CI on exactly one test — `proposal-storage-sqlite.test.ts` › "concurrent reject + gate-decision mutation cannot revive a pending row", at 5118/1 across 381/381 files, every state-gc/migration/improve-lane test green. Verified flake, second occurrence (first was on #738): the test races two real `Worker` threads on one state.db, is untouched by this PR, is CI-green on `main@9f3c8f13` (which contains the identical #738 proposal content, ruling out the only plausible causal story), and passed 5/5 isolated local runs plus a clean 23/23 on the exact CI-head tree. Retriggered via an empty commit (`00745af8`); all five checks green there. If this same WAL test fails a third time it should be investigated as a real race, not retriggered.

---

PR #741, opened 2026-07-31
Fix flag validation, harness detection, and asset ref generation

fwdslsh-dev:
## Summary

This PR addresses several correctness issues across flag parsing, agent harness detection, and asset reference generation:

1. **Unknown flag detection**: Added strict flag validation that rejects undeclared flags instead of silently ignoring them. This catches typos like `--fail-on-flaged` and `--limt` that previously parsed successfully with defaults. Includes a conservative implementation that respects `--` passthrough, skips values after string/enum flags, and handles `--no-` negation correctly.

2. **Harness detection regression**: Fixed `detectHarness` to only report `opencode-sdk` when the `opencode` server binary is actually on PATH. Previously, the SDK import always resolved (since it's a hard dependency), causing every machine to report SDK availability even when the binary wasn't installed, leading to spawn ENOENT errors.

3. **Asset reference generation**: Fixed dangerous-key lint findings and other asset emissions to use canonical, resolvable refs (e.g., `env/prod` instead of `env:prod`). This ensures users can copy refs from findings and use them directly in commands.

4. **Binary download timeout**: Added per-chunk deadline to `streamResponseToFile` to prevent indefinite hangs when an endpoint accepts the request but stalls mid-body. `fetchWithTimeout` only covers time-to-headers.

5. **Task schema validation**: Reconciled lint and parser disagreement on task requirements—lint now correctly requires `version: 2` and `schedule`, while not requiring `enabled` (which the parser defaults to true).

6. **Setup non-interactive guard**: Added TTY check to prevent the interactive setup wizard from blocking forever in piped/CI environments.

7. **Not-found error classification**: Standardized missing resource errors to use `NotFoundError` (exit 1) instead of `UsageError` (exit 2), with better hints that reference the correct commands.

8. **Markdown frontmatter stamping**: Ensured all akm-authored assets include both `type` and `updated` fields, preventing immediate lint failures on newly created assets.

9. **Agent classification fix**: Removed `toolPolicy`-only files from agent classification since the field is never read back—only `tools` determines agent status.

10. **CLI flag alias handling**: Fixed `--auto-fix` / `--fix` to be separate args instead of aliases, since citty's mri integration only supports single-character aliases.

## Test plan

- Added comprehensive test suites for unknown flag detection, harness detection, and not-found error classification
- Updated existing tests to reflect corrected behavior (task schema, asset refs, exit codes)
- All tests pass (`bun test`)
- Lint passes (`bun run lint`)

## Checklist

- [x] Tests pass (`bun test`)
- [x] Docs updated (getting-started guide, workflows reference, stash-makers guide)

https://claude.ai/code/session_01ESAzx67YCzHzg6J1ZSZYPX


itlackey:
@codex review this PR and ensure the repo is ready for release

fwdslsh-dev:
Addressed the Codex review in `f4a78fc` (rebased onto `2e289c8`). Each finding was reproduced against the running CLI before fixing.

**`--no-` on a value flag → internal error (fixed).** Confirmed: `akm search foo --no-limit` exited 70 with `raw.trim is not a function` — mri handed `limit: false` to a string parser. The validator now tracks boolean flag names separately, so `--no-` negates only declared booleans. `--no-limit` is a clean `UNKNOWN_FLAG` (exit 2); `--no-track-usage` and `--no-init` still work.

**`updated:` stamp re-serialized frontmatter (fixed).** Confirmed: a `# rotate quarterly — see runbook` comment was silently dropped. Added `spliceFrontmatterLine`, which inserts the line before the closing fence and leaves every other byte — comments, quoting, key order — untouched. Only a document whose `type` must actually be corrected still takes the re-serialize path. Lint's `--fix` for `missing-updated` now shares that helper rather than keeping its own copy of the splice.

**`enabled` documented as required (fixed).** The task-adapter header and the 0.9.0 adapter spec table now state that `enabled` is optional, defaults to `true`, and must be a boolean when present — matching what the parser and `schemas/akm-task.json` have always done.

**`toolPolicy` classification docs — not stale as of `2e289c8`.** That commit deliberately kept `toolPolicy` as a legacy synonym and wired the renderer to honour it (`tools ?? toolPolicy`), rather than dropping the key. So the docs describing both keys were correct; I reworded them to name `toolPolicy` explicitly as a legacy synonym so the contract and behaviour agree in that direction.

Verified: `tsc` clean, `bun run lint` green, integration 5069/5069. Unit is green except one pre-existing failure on `main` (`goldens-cli-health-tasks`), which is environment-dependent — its expected `remediation` key only appears when a scheduler backend is installed, so it passes in CI and fails on a bare container.

---

PR #745, opened 2026-08-04
0.9.0 final pass: unblock the 0.8→0.9 upgrade, wire adapter validate(), UX and docs honesty

itlackey:
## Summary

A verified review of 0.9.0-rc.15 and the fixes it turned up. Findings were reproduced against the **real published `akm-cli@0.8.14`** installed from npm into an isolated `HOME` with genuine 0.8 state (memories, imported knowledge, a workflow asset, an active workflow run, populated `state.db` + `workflow.db` + `index.db`), then driving the documented upgrade — not read out of the docs.

### Release blocker: fresh 0.8 installs could not migrate

A fresh `akm-cli@0.8.x` install writes `config.json` with **no `configVersion` key at all** — 0.8 stamps it only when a 0.7-era migration did substantive work (`if (changed)`), and a config 0.8 wrote itself has no legacy keys to trigger that. `inspectConfig` read the absent key as `inconsistent`, an unconditional blocker, so `migrate status` reported `blocked` and `migrate apply` refused with exit 78. **Every user who installed 0.8 fresh was unable to upgrade.**

An *absent* `configVersion` on a positively pre-cutover-shaped config is now classified `old`; a present-but-unparseable one still stays `inconsistent`. The transform path needed no change — only the classification was wrong.

Every migration fixture hand-wrote `configVersion: "0.8.0"`, which is why the integration suite never caught it. Added coverage for the shape 0.8 actually produces, plus a negative case.

### `akm-migrate` now derives the 0.9 config

Upgrading required hand-authoring a complete 0.9 config before `apply` would do anything. `bundles`/`defaultBundle` are a pure function of the 0.8 keys, so they are now derived by reusing `migrateConfigSourcesToBundles`. Engine settings are **not** guessed: `profiles.*` / `defaults.llm|agent|improve` are stripped and reported individually in `droppedKeys` by exact dotted path.

Generation is confirm-then-apply. The first `apply` with no `--config` writes a validated starter config and **stops** — no backup run, no sentinel, and `config.json`/`state.db`/`workflow.db` byte-for-byte unchanged. A second `apply` performs the cutover. An explicit `--config` always wins.

### `validate()` is now load-bearing

`validate()` is a REQUIRED interface member implemented by all 11 adapters, and nothing called it. Reproduced consequences:

- **OKF's `missing-ref` never ran.** A bundle with a dangling link reported nothing.
- **The one check the CLI did re-implement had already drifted** — the adapter treats `missing-type` as non-blocking INFO; the CLI copy presented it as a flagged issue with different wording. Divergent behavior, not just risk of it.
- **llm-wiki's four checks were unreachable dead code.**

Adds the single core `ValidateContext` overlay the interface contract calls for and routes lint through each bundle's own adapter. The akm sweep is extracted unchanged — every akm lint test needed only an `await`, **no assertion changes**.

The proposal pre-commit path also runs `validate()`, **advisory only**: it warns and does not reject the write, because the new `resolveRef`'s direct-path fallback disagrees with the legacy resolver on foreign-typed cross-bundle refs. Blocking without reconciling the two resolvers would risk false-positive rejections in a live write path. Documented as such.

`placeNew()` is deferred to 0.10 (D12) — 36 `stashDirFor` call sites across 19 files on the write path, no user-visible benefit today.

### Other fixes

- `akm help migrate 0.9.0` — the command the release notes tell you to run — **failed with exit 78 on a 0.8 config**. `help`/`hints`/bare `akm` now bypass the startup config load.
- `akm setup` dead-ended on an old config. It now refuses cleanly, leaves the file byte-for-byte untouched, and names `akm migrate status`.
- `akm migrate` was `hidden: true` and absent from `--help` entirely.
- `migrate status` on a machine with nothing to migrate reported `blocked`/exit 1; now `not-applicable`/exit 0.
- Post-migration warning when an active run targets a workflow asset that fails 0.9 validation (0.8 heading-format workflows aren't auto-translated).
- Retired hints for `vault`, `task show`, and the retired *flags* (`--background`, `--detect-only`, `--reset-recommended`, `--watch`, `--debounce-ms`).
- `--format text` for `health`/`lint`/`info` dumped raw JSON arrays as single lines (and `info` silently dropped most fields). Now readable, worst-status-first; `json`/`jsonl`/`yaml` unchanged.
- `--format` now works on `akm migrate` instead of warning and printing JSON anyway.

### Docs honesty

Scoped the adapter-refactor claim to what ships — recognition, indexing, presentation and validation are adapter-driven; **placement is not**, and llm-wiki is read-only like OKF. Documented that `logs.db` sits outside backup/restore, that `index.db` is only `quick_check`'d, and that `backup`/`restore`/`storage` live on the separate `akm-migrate` binary. Dropped brittle exact test counts. Inlined a minimal working 0.9 config and an 0.8→0.9 key mapping table.

## Test plan

Counts below are point-in-time at `a9a2109`; run `bun run check` for current figures.

| Gate | Before (rc.15) | After |
| --- | --- | --- |
| unit | 3226 | **3262 / 0 fail** |
| integration | 4838 | **4865 / 0 fail** |
| tsc / lint / build | clean | clean |
| 0.8→0.9 upgrade harness (24 checks) | **3/24** | **24/24** |

End-to-end upgrade verified from a pristine snapshot of a real `akm-cli@0.8.14` install, **with no hand-editing of the config**: config converted to `bundles`/`defaultBundle`, all 21 state migrations applied, `workflow.db` folded away, `index.db` quarantined, refs re-keyed fully-qualified, active workflow run preserved and re-keyed, verified backup written. Both the `--config` and no-`--config` paths confirmed.

Gap closure verified against fixtures captured *before* the change:

```
OKF   before: flagged=1  (missing-type only, CLI wording; dangling link invisible)
      after:  flagged=2  (missing-ref now reported; missing-type now carries the adapter's own INFO wording)
llm-wiki before: 0 findings (all four checks dead)
         after:  uncited-raw, broken-xref, broken-source, missing-description all fire
```

The one changed golden (`okf-conformance`) is legitimate: its fixture always contained a dangling link, and the old `flagged === []` assertion only passed because the check was dead code.

## Checklist

- [x] Tests pass (`bun test`)

## Note for reviewers

`akm migrate` JSON is now **pretty-printed** rather than compact single-line. Semantically identical and JSON parsers are unaffected, but anything grepping `'"status":"blocked"'` will break. STABILITY.md and `docs/reference/cli.md` updated.

The companion onboarding bundle is refreshed in **itlackey/akm-stash#6** — the README quick start currently sends new users to a bundle that is entirely 0.8-era (125 retired command occurrences, ~108 legacy `type:name` refs, and an onboarding workflow that cannot run). Worth landing alongside this.

Copilot's review round (4 comments, all valid) is addressed in `a9a2109` — see [the reply](https://github.com/itlackey/akm/pull/745#issuecomment-5183354115) for what each turned out to be, including one framing I corrected.

### DIFF
--- a/STABILITY.md
+++ b/STABILITY.md
@@ -70,8 +70,8 @@ enumeration of the whole `proposal` noun group.
 | `akm registry list` | Evolving | |
 | `akm registry add` | Evolving | |
 | `akm registry remove` | Evolving | |
-| `akm migrate status` | Internal | Thin forwarder to the standalone `akm-migrate` tool; `hidden: true` in `--help`. |
-| `akm migrate apply` | Internal | Thin forwarder to the standalone `akm-migrate` tool; `hidden: true` in `--help`. |
+| `akm migrate status` | Internal | Forwards to the standalone `akm-migrate` tool; renders its result through the normal `--format` pipeline (not exempt — see below). Listed (not hidden) in `--help`/completions. |
+| `akm migrate apply` | Internal | Forwards to the standalone `akm-migrate` tool; renders its result through the normal `--format` pipeline (not exempt — see below). Listed (not hidden) in `--help`/completions. |
 | `akm config path` | Stable | |
 | `akm config list` | Stable | |
 | `akm config get` | Stable | |
@@ -190,16 +190,23 @@ enumeration of the whole `proposal` noun group.
   `--shape` (`human|agent|summary`) is the output-projection axis (see
   Experimental). A small set of commands is **format-exempt** because their
   output is not a result envelope at all: `completions` (shell script source),
-  child-process passthrough in `env run` / `secret run` / `migrate` (`status`
-  and `apply` both spawn the
-  standalone `akm-migrate` tool), a bare-path payload from `env path`, and
-  document payloads from `help` (bare, `help agents`, and `help migrate`). The
-  set is declared in
+  child-process passthrough in `env run` / `secret run`, a bare-path payload
+  from `env path`, and document payloads from `help` (bare, `help agents`, and
+  `help migrate`). The set is declared in
   `src/output/format-exempt.ts`, and
   passing `--format` to one of them warns rather than silently doing something
   else. Scripted `setup` modes emit a normal format-aware result; interactive
   `setup` is a terminal UI and emits no result document. `agent` leaves
   inherited child streams raw and formats its final result envelope.
+  `migrate status`/`apply` both spawn the standalone `akm-migrate` tool but are
+  NOT in the exempt set: `src/commands/migrate-cli.ts` parses the child's
+  final JSON result line and renders it through the same `output()` pipeline
+  (registered shape `src/output/shapes/migrate.ts`, text renderer
+  `src/output/text/migrate.ts`), so all six `--format` values work on them
+  like any other command. Any progress-event lines the child prints during a
+  real `apply` (content migration, proposal-ref repair) still print verbatim,
+  ahead of the formatted result — those are operational logging, not part of
+  the result envelope.
 
   | Exit code | Meaning |
   | --- | --- |
--- a/STABILITY.md
+++ b/STABILITY.md
@@ -459,6 +459,15 @@ Internal replacement for the one capability nothing else covered.
   a separately published `akm-migrate` package (see Internal above).
 - **0.10 — `--auto-accept` hard error.** It is currently accepted-and-warned;
   see the Improvement loop entry.
+- **0.10 — `BundleAdapter.placeNew()` wiring.** The interface declares
+  `placeNew()` as an optional capability method, and 8 of the built-in
+  adapters already implement it, but nothing in the write path calls it —
+  writes still resolve through AKM's native flat type→directory table.
+  Placement for every existing bundle is already correct today; this is a
+  deliberately sequenced routing change, not unfinished behavior. See
+  [D12](./docs/architecture/specs/0.9.0-decisions.md#d12--bundleadapterplacenew-stays-unwired-until-010)
+  for why it is scoped out of 0.9.0. Nothing user-visible changes in 0.10 for
+  this alone.
 - **1.0 contract freeze** — the `[bundle//]conceptId[#fragment]` ref grammar,
   the supported source model, search behavior, and write-target rules are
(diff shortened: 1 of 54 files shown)


itlackey:
All four review comments were valid. Fixed in a9a2109; full suite green (3262 unit / 4865 integration), tsc/lint/build clean.

**`validate-context.ts:94` — directory treated as a resolved ref.** Confirmed. `existsSync` returns true for directories, so a `pages/foo/` directory with no `pages/foo.md` made a broken xref look resolved and suppressed the diagnostic. Now `statSync(...).isFile()`. Overlay entries are pending file contents by construction, so they still count as files.

**`validate-context.ts:188` — ref containment.** The `..` half was real and is now rejected before joining, along with absolute paths, mirroring the `isWithin` rule the write path already enforces.

One correction for the record: the absolute-path half was **not** actually exploitable. `path.join(root, "/etc/passwd")` yields `<root>/etc/passwd` — still under the root — so absolute refs never escaped. The `..` traversal did. The guard now covers both anyway, so the boundary is stated explicitly rather than resting on a quirk of `path.join`.

Both fixes have regression tests that I verified **fail without the change** (reverted the fix, watched them go red, restored):

```
(fail) resolveRef > a ref naming a DIRECTORY does not count as a resolved target
(fail) resolveRef > rejects refs that escape the bundle root via `..`
```

The absolute-path test passes either way, for the reason above — it pins intent, not a live bug.

**`lint/index.ts:149` — Windows separators.** Confirmed. `path.join` emits `\` on Windows, so `full.includes("/.cache/")` never matched there. Now compares path segments, which also fixes a second bug in the same line the comment hints at: the substring test would have skipped any file under a legitimately-named `registry` directory.

**`lint-format.ts:25` — fix ordering.** Confirmed. `renderStatusEntries` sorts ascending-as-more-urgent, and `true`/`"failed"` both had rank `0`, so a failed fix could sort behind a successful one in the same section — contradicting the comment directly above it. Three states now get three distinct ranks: `failed` (0) < `flagged` (1) < `succeeded` (2), which is what that comment claimed all along.

Re-verified after the hardening that the gap closures this PR is about still hold — OKF still reports `missing-ref`, and llm-wiki still reports `uncited-raw`/`broken-xref`/`broken-source`.

---

PR #779, opened 2026-08-08
feat(workflows): exec units, parallel fan-out defaults, judge redaction, lint polish

itlackey:
## Summary

Closes four gaps left open after the workflow-engine correctness pass in #777.

> **Scope note.** An earlier revision of this PR added `pattern` support to the schema subset, defended by a regex screening analyzer, failed exec units on output overflow, and capped parser diagnostics. All were tripwires — machinery that made runs *more* likely to fail without earning it — and have been removed or narrowed. See [What was deliberately removed](#what-was-deliberately-removed).

### 1. The gate judge no longer bypasses redaction

Every unit outcome is scrubbed by `redactUnitOutcome` before being journaled — but the gate judge dispatched with no `sensitiveValues` and synthetic `runId/stepId/unitId = "gate"`, while its response *is* journaled into the gate row. Since the judge prompt contains the promoted artifact, a secret that reached an artifact could be echoed back and persisted unredacted. This was the only dispatch path writing to the journal without the scrub.

The judge now uses the same redaction contract and carries the real run, step, and gate ids, taken from the same helper that writes the row so the two cannot drift.

**Judge `env` stays absent, deliberately.** `IrGateNode.judge` is an `IrInvocation` (engine/model/timeout/llm) and the decoder rejects unknown keys, so no workflow *can* declare judge env; a step's unit `env` was granted to the work, not the verifier; and an llm judge carrying env fails `env_unsupported` — a gate that could never verify.

**No hashed preimage changed**, so in-flight runs still resume — proven by recomputing a loop-2 `input_hash` and asserting it equals the journaled row's.

### 2. Fan-out has a usable default width

`map.concurrency` and llm engine concurrency both defaulted to `1`, at two independent layers, so the host cap (`min(16, cores−2)`) almost never bound and a `map` over 500 items ran one at a time unless the author opted in twice.

- **map concurrency: 1 → 4.** Not the host cap — a plan frozen on a 32-core CI box would read `32` and mean something different everywhere; 4 is a number an author can reason about, and the host cap remains the safety net.
- **llm engines split by endpoint: loopback 1, remote 4.** Loopback stays serial because that failure is *hard* — a local server holds one loaded model and returns HTTP 500 under concurrent inference (the incident recorded in `AGENTS.md`). Remote failures are soft, retryable 429s. Loopback detection covers `127.0.0.0/8`, `localhost`/`*.localhost`, `::1`, the unspecified addresses, and both IPv4-mapped spellings, purely syntactically (no DNS, so freeze stays deterministic).
- **Three opt-outs**: per-step `concurrency`, per-engine limit, and a `workflow.defaultMapConcurrency` config key that restores the old serial default install-wide.

**Frozen plans are structurally unaffected**: the decoder *requires* an integer width on every map node, so a stored plan cannot inherit a new default. Verified with a 0.9.0-shaped `plan_json` that round-trips byte-identical.

Also fixed: an engine `concurrency` above the decoder's bound used to freeze a plan that then refused to load, and `freeze.ts` recovered source steps by array index — correct only while compile stays 1:1 — now keyed by `stepId`.

### 3. New `exec` (shell) unit kind

Deterministic work — running a test suite, building, invoking a script — previously cost a full LLM or agent dispatch.

```yaml
- id: test
  unit:
    exec:
      command: ["bun", "run", "test:unit"]
      cwd: packages/core
    timeout: "10m"
    retry: { max: 1, on: [non_zero_exit] }
```

- `exec` lives **inside the existing `unit:` bag**, so `timeout`/`retry`/`on_error`/`output`/`env`/`isolation`/`map` all apply unchanged.
- **argv array, no shell form.** The child is spawned directly, so `;`, `|`, `$(…)`, backticks and globs in an argument are inert bytes — the injection class is absent rather than escaped. A pipeline is written `["bash","-lc","a | b"]`, keeping that choice visible in review.
- An exec unit **names no engine**, so an exec-only workflow runs on an install with nothing configured.
- Context arrives as environment (`AKM_RUN_ID`, `AKM_PARAMS`, `AKM_ITEM`/`AKM_ITEM_INDEX`, `AKM_INPUTS`), applied after `env:` bindings so a binding cannot shadow it.
- **Output**: stdout, or exactly one strictly-parsed JSON value when the step declares `output`. Strict parse rather than an embedded-JSON scan, so log noise containing a JSON island is never promoted; a schema miss never re-runs a side-effecting command. stderr never enters the artifact, but a failed unit's stderr **is** journaled as its diagnostic, so `status --units` shows the cause rather than a bare failure reason.
- **Large output does not fail the command.** Output is retained up to a cap and the excess discarded while the reader keeps draining, so the child never blocks on backpressure and keeps its real exit code. A step with no declared schema succeeds with its output marked truncated; only a step whose `output` schema must validate the value fails, since validating a truncated value is meaningless.
- An incomplete capture (pipe error, or drain timeout while the command still exits 0) fails the unit rather than promoting a partial artifact — the exec and agent paths share one classifier so they agree by construction.
- **Non-blocking**: async subprocess and async filesystem checks only, honors `AbortSignal`, kills the process group.
- `cwd` containment is checked at parse time, in the decoder, and again against the resolved base before spawn.
- `akm show` reports an exec step's actual command and no longer claims it uses the workflow's default engine.

**The child's environment is an allowlist.** It starts empty and receives only what a command needs to run — command resolution, home, identity/locale (stdout *is* the artifact, so a missing locale corrupts it), and the Windows variables without which `CreateProcess` fails. This matches the agent-harness convention, and both paths share one helper. Two opt-ins avoid an all-or-nothing cliff: `pass_env: [NAME…]` for per-machine values an env binding cannot express (bindings are committed bundle assets), and `inherit_env: true` for wholesale inheritance.

An allowlist bounds **accidental** exposure and makes the environment surface reviewable. It does not stop a determined attacker, who can already read credentials off disk — that remains the trust model in `docs/guides/run-workflows.md`. Relatedly, **`pass_env` values are deliberately not redacted** from output, because an unlisted name is treated as credential-bearing and scrubbed, which would corrupt a path value. Secrets belong in `env:` bindings.

Context variables are size-checked against **the running platform's** real ceiling before the spawn, so an oversized artifact yields an actionable error instead of a raw `E2BIG` — the check never rejects a spawn that would have succeeded here.

`hashVersion` stays 4 — the exec keys are conditional and each has one encoding per state. Verified byte-identical against a pre-feature worktree.

### 4. Schema and lint polish

- **`allOf`, `anyOf`, `oneOf`, `not` are now enforced** instead of rejected. Evaluation is depth- and budget-bounded and **fails closed** — a truncated evaluation never reports a value valid.
- Unsupported keywords name a working replacement (`const` → single-value `enum`, `$ref` → inline).
- **Lint findings carry line numbers** in both human and JSON output (optional field, so non-workflow sources are byte-identical).

## What was deliberately removed

Applying the rule *no machinery that makes failure more likely without a justified reason*:

- **`pattern` support and its regex screening analyzer — deleted** (`json-schema.ts` 1228 → 605 lines). Supporting author-written regexes from shareable bundles meant either running them unscreened (a measured 4-second synchronous block for a 183-character pattern) or maintaining a soundness analyzer. The analyzer worked, but it rejected legitimate patterns — including the common hand-rolled email regex — to guard a keyword nothing in this repo uses. `pattern` reverts to the plain unsupported-keyword error already shipped in #777.
- **Kill-on-output-overflow — replaced** by drain-and-discard. Killing a passing verbose command over log volume is a tripwire; an OOM of the engine is the worse failure, so the memory bound stayed.
- **Windows-minimum context caps — replaced** by per-platform limits, so a Linux run is no longer failed for a Windows constraint.
- **The parser diagnostic cap — deleted.** It bounded *display* where errors are *produced*, by appending a synthetic error carrying a borrowed line number — which this same branch then rendered as a machine-readable lint finding pointing at a line where nothing is wrong. The input was already bounded (256 steps, 64 params, 1 MiB source).

A follow-up cleanup pass (`ec07e39`, **net −309 lines**) removed further machinery the change did not need: a hand-rolled IPv4/IPv6 address parser replaced by `node:net`, a subprocess result flag nothing read, optional-field gymnastics no caller distinguished, a step's parameters serialized once per *unit* rather than once per step (a wide fan-out retained a distinct copy per unit), a prompt assembled for units that never read one, synchronous filesystem calls on the dispatch path, a diagnostic stored in two fields, an alias constant, a derivable judge identity field, a re-spelled runtime-kind union, a duplicated clip helper, a re-export kept alive by one test import, and documentation restating the same rationale several times.

## Test plan

Run independently on the combined tree after every agent finished:

```
bunx tsc --noEmit          exit 0
bun run lint               exit 0
bun run test:unit          3543 pass / 0 fail  (254 files)
bun run test:integration   5043 pass / 0 fail  (392 files)
```

Plus `node-smoke` under Node 22 and the `AKM_RUN_SLOW_TESTS=1` gated suite, both run locally. Baseline at branch point was 3406 / 4977. Both shrink-only architecture ratchets (function size, import cycle) are green with **untouched baselines** — functions were decomposed rather than rebaselined, and the shared redaction helper was made generic so it imports nothing from its consumers.

Notable coverage: a parent-env variable absent from the child by default and present under each opt-in; a structural pin that *no* name outside allowlist ∪ bindings ∪ `AKM_*` reaches the child; shell metacharacters returned byte-identical; timeout/abort killing child and grandchild; a 12 MiB writer whose unit still succeeds with bounded retention and a truncation marker; a leader exiting 0 while a detached descendant holds stdout, failing rather than promoting a prefix; env values redacted from stdout and stderr; exec units in a `map` fan-out (deterministic peak-in-flight counter and a two-process rendezvous, not wall-clock); worktree isolation against a real git repo; replay reuse proving a completed exec unit is not re-run; and ~40 loopback boundary cases including `127.0.0.1.evil.com` and `0127.0.0.1`.

New tests were verified as real regressions by surgically reverting the file under test.

## Known follow-up, not in this PR

`getDefaultLlmConcurrency` in `src/indexer/indexer.ts:224-235` carries the *same* exact-host loopback bug on the indexing path, and can now import `isLoopbackHost` to converge the two. Left out to keep this PR to the workflow engine.

## Checklist

- [x] Tests pass (`bun test`) — 3543 unit + 5043 integration, 0 failures
- [x] Lint passes (`bun run lint`) — exit 0
- [x] Docs updated — `workflow-schema.md`, `author-workflows.md`, `run-workflows.md`, `configuration.md`, `cli.md`, and `CHANGELOG.md`

https://claude.ai/code/session_01AkFoKnRrLrbrL7RDFkEcm9


itlackey:
@codex do a deep review of theae changes to find all defects, gaps, misalignments

itlackey:
All 7 review findings are addressed in `8b51d01`. I assessed each as legitimate — none were dismissed.

**Verified after the fixes:** `tsc` clean, `lint` exit 0, **3558 unit / 5037 integration, 0 failures**, both shrink-only architecture ratchets green with untouched baselines, plus `node-smoke` (Node 22) and the `AKM_RUN_SLOW_TESTS=1` suite run locally. Every new test was checked to fail without its fix.

**P1 — unbounded exec output capture.** A breach now kills the process group and fails the unit (`exec_output_limit`, 8 MiB/pipe) rather than truncating. Truncating was rejected because stdout *is* the artifact: a silently shortened one flows into the gate judge and the next step's inputs while every status surface still reports success. "Truncate and keep running" also isn't actually available — once the reader stops draining, the child blocks on its next write and it degenerates into a hang until the wall timeout. The cap went into the subprocess primitive as **opt-in**, so existing callers are byte-for-byte unchanged; the agent spawn path is also unbounded and was deliberately *not* opted in mid-PR — recommend a follow-up.

**P1 — sequential ambiguous regex groups.** Confirmed and worse than described. Against the previous screen, `^(a|aa)…(a|aa)b$` × 30 is 183 chars, **accepted**, and blocks for **4.15 s**; the 40-group variant is still under the cap and extrapolates to hours.

| groups | 24 | 28 | 30 |
|---|---|---|---|
| match | 139 ms | 1.30 s | 4.15 s |

A runtime bound was ruled out after verification, not assumption: a wall-clock timeout cannot preempt a synchronous `RegExp.exec`, and moving the match off-thread would make validation asynchronous and wall-clock-dependent — unacceptable when it feeds gate decisions. The bound is therefore established before matching. The shape blacklists were replaced with structural rules: alternatives must be provably disjoint **whether or not the group is quantified** (the actual hole), and a varying-length atom must end disjoint from what can follow. Lookarounds were being modelled as consuming — which could fool the screen into believing a boundary was forced — and are now zero-width; backreferences are rejected outright. A choice-point budget backs this as a failsafe. Empirical check: 30k generated patterns, worst match 0.28 ms.

*One deliberate collateral rejection:* the hand-rolled email regex `^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$` — its class contains its own `.` delimiter, so nothing forces which dot ends the repeat. That's textbook polynomial ReDoS, so it's a true positive; docs name it with the accepted rewrite `[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)*`.

**P2 — incomplete pipe capture.** The exec and agent paths now share one classifier, so they agree by construction rather than convention; a failed or timed-out drain fails the unit instead of promoting a prefix.

**P2 — stderr diagnostics.** The already-redacted diagnostic is journaled to the column `status --units` already reads, and appended to the run summary. A resume rebuilds a byte-identical summary from the journal.

**P2 — context in environment variables.** Bounded and rejected before the spawn with an actionable error, over a file transport — the latter would make the `AKM_INPUTS` contract conditional on data size (sometimes JSON, sometimes a filename), forcing every command to handle both. Bounds are pinned to the *smallest* platform ceiling (Windows' 32 767-char per-variable limit) so a workflow can't pass on Linux and die on a Windows runner. `AKM_PARAMS` and `AKM_ITEM` are covered too.

**P2 — loopback range.** Now a purely syntactic classifier over all of `127.0.0.0/8`, `localhost`/`*.localhost`, `::1` in any spelling, IPv4-mapped forms, and the unspecified addresses, with ~41 boundary cases pinned including `127.0.0.1.evil.com` and `0127.0.0.1`. No DNS, to keep freeze deterministic. It also fixed `new URL("localhost:1234")` yielding an empty hostname that previously fell through to *remote*.

**P2 — show projection.** `exec` steps now report their command and no longer claim a default engine. Argv is shown **unclipped** — clipping would reintroduce the exact defect this finding is about, and argv is safe to display since the format has no substitution language, so every byte is already literal in the asset being rendered. Renderer goldens regenerated byte-identical.

`hashVersion` stays 4 and no hashed preimage changed — pinned golden hashes captured before this work still pass, so in-flight runs neither re-dispatch nor diverge.

**One finding outside this PR's scope, reported not fixed:** `getDefaultLlmConcurrency` in `src/indexer/indexer.ts:224-235` carries the *identical* exact-host loopback bug on the indexing path. It can now import `isLoopbackHost` to converge the two. Happy to fix it here or separately.

---

itlackey:
Trimmed in `4728dc3`, applying the rule that nothing should make a run more likely to fail without a justified reason. The PR body and title are updated — they previously claimed `pattern` support and kill-on-overflow, both now false.

**Removed — `pattern` support and its regex screening analyzer** (`json-schema.ts` 1228 → 605 lines; zero references to any analyzer symbol remain). Supporting author-written regexes meant either running them unscreened or maintaining a soundness analyzer, and the analyzer rejected legitimate patterns — including the common hand-rolled email regex — to guard a keyword nothing in this repo uses. `pattern` reverts to the plain unsupported-keyword error already shipped in #777. The `format` hint that read "use `pattern`" was corrected, since it would otherwise have pointed at another unsupported keyword.

**Narrowed — output overflow no longer kills the command.** The reader keeps draining and discards past the retention cap, so the child never blocks on backpressure and keeps its real exit code. A step with no declared schema now succeeds with its output marked truncated; only a step whose `output` schema must validate the value still fails, since validating a truncated value is meaningless. The memory bound stayed — an OOM of the engine is the worse failure — and retention is proven to stop exactly at the cap against 12 MiB and 20 MiB writers.

**Narrowed — context size is checked against the running platform's ceiling**, not the smallest platform's, so a Linux run is no longer failed for a Windows constraint. The guard now only fires where the spawn was going to fail anyway, turning a raw `E2BIG` into an actionable error.

Kept, because the alternative failure is worse in each case: failing an incomplete capture rather than promoting a partial artifact as success, the env allowlist (two opt-outs), and the judge redaction fix.

Net: this revision removes two failure modes an earlier revision introduced and narrows a third, while keeping the fixes that prevent silent data loss.

Verified after the trim: `tsc` clean, `lint` exit 0, **3546 unit / 5043 integration, 0 failures**, both shrink-only ratchets green with untouched baselines, plus `node-smoke` and the `AKM_RUN_SLOW_TESTS=1` suite locally. CI is green on this head.

---

PR #822, opened 2026-08-25
release: merge 0.9.2 into main

### LINKED ISSUE #753: Write-path index freshness: remember and bundle add leave new assets unsearchable until a manual `akm index`
**Problem**

Fresh writes are not reliably visible to `search`/`curate` until someone runs `akm index`. This breaks the compound-engineering pattern (#754) where session N writes knowledge (`akm remember`) and session N+1 discovers it via `akm curate` — the failure is *silent* (stale results, not an error). It also affects `akm bundle add`: a newly installed bundle's assets aren't searchable until the user separately runs `akm index` (the README quick start documents this two-step dance explicitly).

**Confirmed by code inspection (2026-08-06)** — this is no longer an open question; the mechanism map is:

- **Write-path fast indexing already exists**: `src/indexer/index-written-assets.ts` (`indexWrittenAssets`) upserts FTS entries for specific written files under the index-writer lease, fail-open (any error degrades to "appears after the next full `akm index`"). It deliberately **skips embeddings** and the other LLM passes — its own doc comment defers those to `ensureIndex`/`akm index`.
- **It's already called by**: `akm import` (`src/commands/read/knowledge.ts` — "Write-path indexing: the asset is searchable immediately"), `akm clone` (`src/commands/sources/source-clone.ts` — targeted index update with a warning on failure), improve's extract (`src/commands/improve/extract.ts`), and proposal writes (`src/commands/proposal/repository.ts`).
- **It's NOT called by `akm remember`** (`src/commands/remember.ts` → `writeAssetToSource` writes the file and takes the mutation lease, but nothing indexes it) — ironic, since remember is the flagship in-session capture command.
- **It's NOT called by `akm bundle add`** (`src/commands/sources/installed-stashes.ts` / `add-cli.ts`) — user-facing messages there say "run `akm index` to refresh the search index".
- **The read path does not self-heal**: only `show` runs a read preflight (`ensurePrimaryIndexForRead`, `show.ts:111`), and it calls `ensureIndex` in *background* mode, which checks `indexCanServeStash` — DB exists, has rows, covers this stash — and explicitly serves "at worst content-stale" results. The mtime-based freshness check (`isIndexStale` → `hasNewerIndexableFiles`) exists but is only used in *blocking* mode by improve's planning preflight. `search`/`curate` don't run a freshness preflight at all.

**Proposed solution**

1. **`akm remember` calls `indexWrittenAssets`** on the file it just wrote, exactly as `import`/`clone`/proposal writes already do. This is the smallest change with the highest value: memories become FTS-searchable immediately, using an existing, tested, fail-open mechanism.
2. **`akm bundle add` (and `bundle update`) indexes the new bundle immediately** after a successful sync, instead of printing "run `akm index`". For a whole bundle the single-file fast path may be the wrong tool — a scoped incremental `akm index` run covering just the new bundle's root (the machinery for incremental per-file re-extraction already exists in `indexer.ts`) is likely the right shape, and unlike the fast path it can include embeddings since bundle-add is already a long-running, network-bound operation where a few extra seconds is acceptable. Failure stays fail-open with the current "run `akm index`" message as the degraded path.
3. **`akm import` — verify and pin, no change expected**: import already calls `indexWrittenAssets`. Add/keep a regression test asserting an imported doc is FTS-searchable immediately with no intervening `akm index`, so the guarantee is pinned rather than incidental.
4. **Document the embeddings caveat as the contract**: write-path-indexed assets are **FTS-fresh immediately, embedding-fresh after the next full index** (hybrid search will still surface them lexically). State this in `docs/reference/workflows.md` and `docs/guides/improvement-loop.md` so workflow authors know exactly what "immediately searchable" means. If semantic-freshness-on-write is ever wanted, that's a follow-up that must handle provider latency/failure on the write path — out of scope here.
5. **Optional follow-up — read-path staleness**: background `ensureIndex` already supports `hydrateSources: false` (rebuild from local content only, never clone/pull at query time), so adding the `hasNewerIndexableFiles` mtime check to the read path would not violate the "query time never fetches" rule (spec §14.3 / D11). The tradeoff is a per-read mtime scan; measure before adopting. With items 1-2 in place this becomes a belt-and-braces safety net rather than the primary mechanism.

**Implementation details**

- `remember` → `indexWrittenAssets`: mirror the call shape in `knowledge.ts` (write file → fast-index → fail-open warning on error). The write already happens through `writeAssetToSource`, which knows the resolved target source root — that's the `stashDir` argument the fast path needs, plus the bundle id for provenance (`deriveInstallations` handles this inside the helper).
- `bundle add` → scoped index: after `syncFromRef` succeeds and config/lock are published (the ordering comment at `installed-stashes.ts:449-454` already anticipates "config/lock publication and reindexing both succeed" as the completion condition), trigger the incremental index scoped to the new bundle before cleaning up the old root. Respect `index-writer-lock`; on failure, keep today's "run `akm index`" message.
- Tests (`tests/` unit + `tests/integration/`): (a) `remember` then `search` finds the memory with no `akm index` between; (b) `bundle add` of a fixture bundle then `search` finds its assets; (c) `import` then `search` (pins existing behavior); (d) a workflow-engine test (`tests/workflows/`) simulating step A `remember` → step B `search`, which is the actual #754 dependency.
- Note for #747 (MCP server): a long-lived server must re-check freshness per request; it should inherit whatever contract lands here rather than inventing its own.
- Note for #748 (chunked embeddings): the chunked schema must keep `indexWrittenAssets` cheap for single-file updates — re-chunking one changed file must not touch other files' rows.

**Alternatives considered**

- Require every workflow/user to call `akm index` after writes (status quo) — pushes a silent-failure footgun onto every workflow author and contradicts the fact that import/clone/proposal writes already got write-path indexing precisely to avoid this.
- Full `akm index` after every `remember` — too slow for an in-session capture command; the scoped fast path exists for exactly this reason.
- Embedding generation on the write path — rejected for now: adds provider latency and a new failure mode to `remember`; FTS-immediate + embedding-at-next-index is an honest, documentable contract.

### LINKED ISSUE #772: Decide how to surface the targeted-write semantic-search visibility gap
## Problem

With semantic search enabled, an asset written through the write-path indexer (`akm remember`, proposal accept, `improve extract`, `source clone`) becomes keyword-searchable immediately but has no vector embedding until the next full/incremental `akm index` run. During that window a semantic/vector query will not surface the just-written asset, and nothing in the CLI output or search response tells the caller that a specific hit (or miss) is due to a pending embedding rather than genuine irrelevance. An agent or workflow step that writes a memory and immediately relies on semantic search/curate to find it again can silently miss it.

## Evidence

- `src/indexer/index-written-assets.ts:5-22` — module doc states embeddings, index-time LLM passes, graph extraction, `builtAt`, and the walk cache are "deliberately untouched — the next full run heals them."
- `src/indexer/index-written-assets.ts:134-170` — the per-file write loop calls `upsertEntry` and (once) `rebuildFts(db, { incremental: true })` at :170; no embedding-generation call exists anywhere in the file.
- `src/indexer/index-written-assets.ts:51` — `WRITE_PATH_INDEX_BUSY_TIMEOUT_MS = 5_000`, documented as a hard latency budget for `akm remember` so it "must not hang... behind a running full reindex" — the module is explicitly optimized for low write latency, which is in tension with adding synchronous embedding generation to it.
- `src/commands/read/knowledge.ts:796,801` — `akm remember`'s write path calls `indexWrittenAssets` after every write with no follow-up embedding step or caller-visible signal.
- `src/commands/read/knowledge.ts:808-830` — the write response already carries an advisory `hint` field (used today only for stash placement-convention advice), the closest existing precedent for surfacing an advisory to the writer.
- `src/indexer/search/db-search.ts:181,259,271,300` — search responses carry a global `tip` advisory field for index-state issues (missing/empty index), but nothing keyed to a specific just-written, not-yet-embedded row.
- `src/indexer/search/semantic-status.ts:31-39` — `SemanticSearchStatus` is a single global status (`pending`/`ready-js`/`ready-vec`/`blocked`) with no per-entry or per-write embedding state.
- `docs/architecture/testing/manual-testing-checklist.md:2965-2968` — the 0.9.0 release triage recorded this explicitly: "Remaining, accepted by design: targeted-write indexing skips embeddings on purpose... whether to surface that gap at write time is a 0.10 placement-era product decision, not a regression."

## Proposed fix

This is a product decision, not a bug fix. Three options, to be decided and recorded:

1. **Leave as-is, document the window.** Cheapest option. State the limitation somewhere a workflow author or `akm remember` caller would actually see (not only the internal module doc), e.g. in the `akm remember`/`akm workflow` reference docs. Leaves semantic-only consumers silently missing fresh writes.
2. **Surface it.** Reuse the existing `hint`/`tip` advisory pattern: add a field to the `akm remember` response (e.g. `embeddingPending: true`) and/or flag not-yet-embedded rows in semantic search results, so a miss is explainable instead of silent. Smallest change consistent with the module's fail-open, low-latency design.
3. **Embed on the write path.** Call the embedding generator inside `indexWrittenAssets`, paying the latency the module currently avoids (works against the documented 5s write-path budget and the module's fail-open design center).

Recommendation: option 2. It doesn't add embedding latency to every write (preserving the module's stated design goal), and it reuses a pattern (`hint`) the write path already has rather than inventing a new one. Option 3 conflicts with the write path's fail-open/low-latency charter. Option 1 is cheapest but leaves the miss silent for any semantic-only consumer (e.g. an agent chaining `remember` then `curate` in the same workflow run).

## Acceptance

- [ ] One of the three options above is chosen and the decision is recorded (e.g. an ADR entry or a decision note under `docs/architecture/`).
- [ ] If option 2 or 3: implemented with a regression test asserting the new visible behavior (e.g. write response carries the pending flag, or a semantic query finds the asset within the write-path's existing latency budget).
- [ ] If option 1: the limitation is documented where a `akm remember`/workflow-authoring user would see it, not only in the module's internal doc comment.
- [ ] `docs/architecture/testing/manual-testing-checklist.md` section 24.2's Semantic-row "Remaining, accepted by design" note is updated or removed once the decision lands.

## Context

Raised in the 0.9.0 release triage (2026-08-06); recorded in `docs/architecture/testing/manual-testing-checklist.md` section 24.2 under the Semantic row as "accepted by design... not a regression," with no formal waiver/expiry attached (unlike the other open §24.2 rows) — the triage instead scheduled the write-time-visibility decision itself for 0.10's placement/adapter-write era. Mitigation that made this acceptable to defer: `indexWrittenAssets` is fail-open and self-heals via the next full or incremental `akm index`, so the gap is a temporary visibility window on semantic-only queries, not data loss — keyword/FTS search already sees the asset immediately.

### LINKED ISSUE #812: extract: unparseable LLM JSON is silently swallowed — raw response discarded, no retry, no top-level warning
## Summary

When the extract LLM returns something `parseExtractPayload()` cannot parse, the run reports success with zero candidates and the only diagnostic is the fixed string `"LLM response was not parseable JSON"`. The raw response is discarded, so there is no way to tell whether the model emitted prose, a truncated object, malformed JSON, or an error body — and there is no retry. Session extraction then silently produces nothing, run after run.

## Environment

- akm CLI 0.9.1, Linux
- Invoked via the akm Claude Code plugin's `SessionEnd` → `extract-session` hook
- `defaults.llmEngine = "default"` → local vLLM, `supportsJsonSchema: false`
- Endpoint verified reachable (`GET /v1/models` → 200), so the call is reaching the model

## Actual

`~/.local/state/akm-claude/extract.log`:

```json
{
  "schemaVersion": 1,
  "ok": true,
  "shape": "extract-result",
  "type": "claude-code",
  "sessionsProcessed": 1,
  "sessionsSkipped": 0,
  "candidatesCreated": 0,
  "proposals": [],
  "sessions": [
    {
      "sessionId": "d517f795-92f0-4cc5-b40b-495fa5627ea6",
      "harness": "claude-code",
      "candidateCount": 0,
      "proposalIds": [],
      "rationaleIfEmpty": "LLM response was not parseable JSON",
      "preFilter": { "inputCount": 12, "outputCount": 9, "truncatedCount": 0 },
      "warnings": [],
      "contentHash": "825fbb182199d8df4a68bc90524ba7cb10801798f0bbf779b4664ec3b8123169"
    }
  ],
  "warnings": [],
  "durationMs": 9885
}
```

The session was read and pre-filtered correctly (12 in, 9 out), so everything up to the LLM call worked. `ok: true` and an empty `warnings` array mean nothing upstream — a health check, a nightly `improve` pass — registers this as a failure.

## Cause

`src/commands/improve/extract-prompt.ts`, `parseExtractPayload()`. The function already tolerates prose preamble/postamble by slicing the first balanced top-level object, which is good, but both failure paths drop `stdout` on the floor:

```ts
const start = stdout.indexOf("{");
const end = stdout.lastIndexOf("}");
if (start === -1 || end <= start) {
  return { candidates: [], rationale_if_empty: `LLM response was not parseable JSON` };
}
try {
  parsed = JSON.parse(stdout.slice(start, end + 1));
} catch {
  return { candidates: [], rationale_if_empty: `LLM response was not parseable JSON` };
}
```

Nothing distinguishes "no brace found at all" from "braces found but the slice did not parse", and the response body is never retained, logged, or surfaced.

## Impact

On an engine with `supportsJsonSchema: false`, free-form output is the expected failure mode rather than an exceptional one, so this is not a rare path — it is the steady state until someone happens to read `extract.log`. Durable knowledge capture stops working with no visible signal.

## Suggested fix

1. Retain evidence: include a truncated preview (say the first 500 chars) and the response length in `rationale_if_empty`, or log the full body behind `--verbose` / a debug env var. Without the raw text the failure is not diagnosable from the outside.
2. Distinguish the two branches — "no JSON object found in response" vs "extracted object failed to parse" point at different problems.
3. Consider one repair retry when the engine has `supportsJsonSchema: false`, re-prompting for JSON only. That is where the failure concentrates.
4. Surface the parse failure in the run's top-level `warnings` so `ok: true` with zero candidates is not indistinguishable from a genuinely empty session.

### LINKED ISSUE #819: Retrieval ceiling: conjunctive-AND FTS + unindexed body prose cap the payoff of any engagement gain
## Summary

Two akm-side properties cap how much value any plugin engagement can deliver, independent of *when* the model calls akm:

1. **FTS is a conjunctive AND with no stopword removal**, so sentence-shaped queries zero-hit while keyword queries rank 1.
2. **Body prose is never indexed** — only name / description / tags / searchHints / headings.

Both were established during benchmark-library authoring; this issue records them as tuning targets now that engagement is being measured.

## Why it matters now

The first real A/B (`akm-bench` `claude/harbor-akm-agent-p0`, `results/harbor/2026-08-24/`) shows akm helps where retrieval is the only path — eval slice paired delta **+0.228 [0.070, 0.404]**, retrieval-family tasks **0.33 → 0.67**. But engagement is only 28%, and every *additional* engagement we win by fixing the plugin's trigger wording (itlackey/akm-plugins#94) only pays off if the resulting query actually retrieves.

In other words: query shaping is the multiplier on every engagement fix.

## Evidence / prior measurements

- Question-form `searchHints` took a benchmark library from **0/8 to 7/8** retrievable — i.e. the authoring rule, not the ranker, decided the outcome.
- akm-eval's LongMemEval work applies a deterministic stopword strip before queries reach akm's FTS specifically to work around the conjunctive-AND behaviour, and records a `zeroHitQueries` / `retrievalQueryCount` counter in `result.json.metadata` because zero-hit rates on natural-language queries are high enough to change how a score should be read.

## Suggested directions

- Make akm tolerant of sentence-shaped input (stopword strip / OR-fallback / progressive relaxation) rather than requiring every caller to pre-shape queries. Each caller currently reimplements this.
- If the conjunctive-AND behaviour is deliberate, surface it: a zero-hit result could say *why* (e.g. "3 of 7 terms matched nothing") rather than returning an empty list, so callers and models can retry usefully.
- Document the "body prose is never indexed" rule prominently in the authoring guidance, since it decides retrieval outcomes more than ranking does.

## Related

- itlackey/akm-plugins#94 — engagement trigger wording (the larger effect)
- itlackey/akm-plugins#95 — tool descriptions vs built-ins
- `akm/docs/plans/benchmark-tuning-findings.md` — full findings and the train/eval tuning protocol

### LINKED ISSUE #820: index --clean leaves stale FTS records for renamed or removed assets
## Summary

`akm index --clean` and `akm index --full --clean` can leave stale full-text-search (FTS) records after an asset is renamed or removed. `akm search` continues to return the old ref even though `akm show` correctly reports that its source file is missing.

## Impact

Search results can direct agents to invalid assets after a rename migration, creating false confidence that deprecated references still exist and wasting follow-up work.

## Reproduction

1. In a configured writable bundle, rename or remove an indexed asset. Example: rename `knowledge/printmd/preview-server-usage.md` to a new `gutterpress`-named path.
2. Confirm the old file/path is absent on disk.
3. Run:

   ```sh
   akm index --full --clean
   akm search "printmd"
   akm show knowledge/printmd/preview-server-usage
   ```

4. Observe:
   - `akm search` still returns `knowledge/printmd/preview-server-usage`.
   - `akm show` fails with `ASSET_NOT_FOUND`, stating the indexed file is missing/unreadable.
   - A direct filesystem scan has no `printmd`/old-path match.

The behavior was also observed after the incremental form, `akm index --clean`.

## Suspected root cause

This is an inference from the externally visible behavior: the clean/full index flow reconciles the asset lookup enough for `show` to reject the old source path, but does not delete or rebuild the corresponding FTS/name rows. The stale document appears to outlive its canonical asset record, likely because cleanup is keyed to one table/path representation and is not cascaded to every search index table.

## Expected behavior

After `akm index --clean` (and especially `--full --clean`), searches must not return entries whose source file no longer exists. The new asset path should be searchable; the old ref should be absent from both `search` and `show`.

## Proposed remediation

- Make index cleanup remove stale rows from every index/search table, including FTS and name/path lookup tables.
- For `--full`, rebuild each source's index atomically from the current discovered asset set, rather than retaining prior rows.
- Use a stable source identity plus canonical relative path when reconciling renames, and delete old identities before inserting replacement rows.
- Add a post-index invariant: every FTS result must resolve through `show`.

## Regression coverage

Add integration tests for both the default bundle and a configured non-default source:

1. Index an asset.
2. Rename and remove the old path.
3. Run `index --clean` and `index --full --clean`.
4. Assert the old query/ref returns no result, the new ref resolves, and each search result's `show` succeeds.

### LINKED ISSUE #846: task sync: reconcile is scoped by bundle NAME, so an isolated bundle can attempt to remove another bundle's real scheduler entries
## Summary

`akm task sync` run from a **fully isolated** bundle attempted to remove a scheduled task belonging to a **different, real bundle**. It failed closed only because the target entry happened to lack a native fingerprint. Had that fingerprint existed, a production cron job would very likely have been deleted.

Found independently by two agents during manual-test design against `v0.9.2-alpha.5` (`fb7bef3f`).

## What happened

An isolated test bundle — `AKM_BUNDLE_DIR`, `AKM_CONFIG_DIR`, `AKM_DATA_DIR`, `AKM_CACHE_DIR`, `AKM_STATE_DIR` all pointed at fresh temp dirs, **zero task ids in common** with the machine's real schedules — ran `akm task sync`. Instead of confining itself to that bundle, sync inspected the machine's real crontab and tried to reconcile against it, specifically attempting to touch `akm-dogfood-091-capture`, which belongs to an unrelated bundle.

It aborted with `RESOURCE_ALREADY_EXISTS` — "Installed scheduler binding … has no exact native fingerprint; refusing removal" (`src/tasks/scheduler-sync.ts` ~:243). Fail-closed, no partial write; the crontab was verified byte-identical before and after.

**The guard that saved it is incidental, not a scoping decision.** It fires on a missing fingerprint, not on "this binding belongs to someone else."

## The mechanism

Removal *is* nominally scoped (`scheduler-sync.ts:178`):

```ts
const scopedInstalled = coherentInput.installed.filter((entry) => belongsToBundle(entry, coherentInput));
```

and (`:715`):

```ts
function belongsToBundle(entry: InstalledSchedulerBinding, input: SchedulerSyncPlanInput): boolean {
  if (entry.target === input.bundleName || entry.target === input.bundleTarget) return true;
  return false;
}
```

with `bundleName` sourced from `bundle.source.name` (`src/commands/tasks/tasks.ts:403`). So the scope key is a bundle **name**, not the bundle's path or identity.

That makes the isolation the env vars provide incomplete: `AKM_BUNDLE_DIR` relocates the bundle's *files*, but the OS scheduler is a single global namespace, and the only thing separating one bundle's bindings from another's is a name string that an isolated bundle can easily resolve to the same value as a real one.

Everything downstream then treats a real, foreign binding as in-scope drift, and `removed` (`:227`) is computed as "installed-and-in-scope minus desired" — i.e. every real schedule the test bundle doesn't declare.

## Why this is serious

- **The blast radius is a user's real scheduled jobs**, not test state.
- It is reachable from what a careful user would consider a fully sandboxed setup — pinning all five `AKM_*` dirs is the documented isolation pattern and is what every hermetic test in this repo does.
- The failure mode is silent-until-destructive: nothing warns that sync is looking outside the bundle.
- `akm task add` does *not* have this problem — it only touches its own task's binding (confirmed by reading `akmTasksAdd`). The asymmetry makes `sync` surprising.

## Reproduction (SAFE — read the warning)

**Do not run this on a machine with real akm schedules you care about.** Back up first: `crontab -l > backup.txt`, restore with `crontab backup.txt`.

```sh
B=$(mktemp -d); mkdir -p "$B"/bundle/tasks "$B"/{config,data,cache,state}
printf '{"configVersion":"0.9.0","semanticSearchMode":"off","registries":[]}\n' > "$B/config/config.json"
env AKM_BUNDLE_DIR="$B/bundle" AKM_CONFIG_DIR="$B/config" AKM_DATA_DIR="$B/data" \
    AKM_CACHE_DIR="$B/cache" AKM_STATE_DIR="$B/state" \
  akm task sync
```

Observed on a machine with 11 real akm cron entries: sync reached for `akm-dogfood-091-capture` (a different bundle's task) and aborted on the fingerprint guard.

## Scope of THIS issue

**Narrowed:** this issue is now only the correctness fix — scope the reconcile by something that cannot collide.

The dry-run/preview affordance originally proposed here has been split into **#849** and assigned to 0.10.0, so 0.9.2 carries only the small, mechanical change that closes the data-loss risk.

## The fix

Scope `belongsToBundle` (`src/tasks/scheduler-sync.ts:715`) by the bundle's **resolved path** (or a stable path-derived id) rather than its display name:

```ts
function belongsToBundle(entry: InstalledSchedulerBinding, input: SchedulerSyncPlanInput): boolean {
  if (entry.target === input.bundleName || entry.target === input.bundleTarget) return true;
  return false;
}
```

A display name derived from a directory basename is not an identity — two unrelated bundles legitimately share one. The resolved filesystem path is the identity that actually distinguishes them, and it is already available at the call site.

Additionally, `sync` should refuse to remove a binding whose owning bundle path differs from the bundle it was invoked from, and say so — rather than depending on the incidental fingerprint check to fail closed.

## Related

- Discovered while designing manual tests for the #844 task/workflow refactor.
- `akm task` has no `remove` subcommand; the documented removal path is "delete the `.yml` and run `task sync`" — which routes users straight through this command.

### LINKED ISSUE #847: workflow abandon: a blocked run becomes permanently unresumable — abandon leaves step status blocked while run status goes failed
## Summary

A workflow run that is **blocked** and then **abandoned** can never be resumed. `resume` hard-fails with a corrupt-spine error, permanently stranding the run — directly contradicting `akm workflow abandon --help`, which states "resume can reopen it".

Reproduced twice independently against `v0.9.2-alpha.5` (`fb7bef3f`) during manual-test design.

## Reproduction

1. Start a workflow run that blocks on a judge/verification gate — e.g. a gate whose judge endpoint is unreachable, so the run lands in `blocked` rather than failing outright.
2. `akm workflow abandon <runId>` — succeeds.
3. `akm workflow resume <runId>` — fails:

```
has a corrupt durable step spine... Refusing to mutate state that disagrees with its frozen plan
INVALID_JSON_ARGUMENT, exit 2
```

The run is now unresumable and unabandonable-further: a terminal dead end.

## Cause

`abandon` sets the **run-level** status to `failed` but leaves the **step-level** status at `blocked`. `resume` then validates run state against the frozen plan, finds run-status and step-status disagreeing, and refuses to touch it — the integrity check is correct; the state it is handed is not.

So `abandon` is writing a partial state transition: it moves the run but not the step it is blocked on.

## Isolating evidence

The ordinary path round-trips fine. A run that is **active (paused)** — not blocked — survives `abandon` → `resume` and reopens normally. Only the blocked case strands. That isolates the defect to `abandon`'s handling of a blocked step rather than to `resume` or to abandon generally.

## Why it matters

`blocked` is not an exotic state — it is what a run enters whenever a verification gate cannot reach its judge, which is exactly the situation a user hits with a misconfigured or temporarily-down judge engine. The natural recovery ("give up on this run, fix the config, reopen it") is the path that bricks it. And `abandon --help` explicitly promises the reopen works.

Practical impact: the user's only recourse is to start a fresh run, losing the abandoned run's history and identity.

## Expected

Either:
- `abandon` transitions the blocked step to a terminal state consistent with the run-level `failed`, so `resume` can legitimately reopen both; or
- `resume` knows how to reopen a run whose steps were left blocked by an abandon, treating that specific shape as recoverable rather than corrupt; or
- if a blocked run genuinely cannot be reopened by design, `abandon` should say so at abandon time and `abandon --help` should stop promising otherwise.

The first is the behavior the help text already advertises.

## Related

Found while designing manual lifecycle tests for the #844 task/workflow refactor. The corresponding test (blocked → abandon → resume) is in the manual-test suite being assembled, with the active-paused contrast case alongside it.

itlackey:
## Summary

- promote `release/0.9.2` into `main`
- cut the package from `0.9.2-alpha.5` to stable `0.9.2`
- include the release-line fixes, including cross-bundle scheduler isolation (#846) and blocked-run abandon/resume recovery (#847)
- ship both the 0.9.1 → 0.9.2 migration guide and the terminal release note used by `akm help migrate 0.9.2`

## Release readiness

- changelog cut as `0.9.2` dated 2026-08-29
- [migration guide](https://github.com/itlackey/akm/blob/release/0.9.2/docs/migration/v0.9.1-to-v0.9.2.md)
- [shipped release note](https://github.com/itlackey/akm/blob/release/0.9.2/docs/migration/release-notes/0.9.2.md)
- active workflow documentation now distinguishes the durable-v4 family from its current executable `irVersion: 5`
- manual task/workflow QA reflects the fixed #846 and #847 expectations

## Rebase integrity

- release line rebased onto `main` at `9473ccf2`
- pre-rebase release tip: `927d16eceba80913d727964edddd8c5e9bbf2ea0`
- rebased release tip: `cab5f29fc99f895c0cc406665931c269bbdd93d0`
- the original 40 release commits were patch-identical under `git range-diff`, and the rebased tree matched the synthetic merge tree
- final stable candidate head: `66cca9990db77c12a72ad0ac9e6f02beaf9efedc`

## Verification

Local verification used the CI-pinned Bun 1.3.14 toolchain:

- lint, TypeScript, generated-schema, documentation-example, and terminology gates: pass
- unit: 4,320 pass, 0 skip, 0 fail across 319 files
- integration: 5,892 pass, 53 expected skips, 0 fail across 443 files
- focused workflow release, migration-help, frozen-plan, and plugin-version regressions: 30 pass, 0 fail

PR checks at `66cca999` are green: 12 passed, 0 failed, and the release-candidate-evidence job was skipped by design. Exact release-candidate evidence is intentionally collected after merge against the immutable merge commit (or its candidate tag), not against a moving PR head.

Closes #753
Closes #772
Closes #819
Closes #820
Closes #812
Closes #846
Closes #847


itlackey:
## 0.9.2 exact-SHA release-candidate evidence

- Candidate merge commit: `31046cc4d7c3ea656b55268e7db27bbb6f9e0965`
- Gated CI run: https://github.com/itlackey/akm/actions/runs/33242292032
- Trigger: manual `workflow_dispatch` with the full `candidate_sha` and `gated_suite=all`
- Result: semantic search, Docker install, and native schedulers on Linux, macOS, and Windows all succeeded; `Gated / Release Candidate Evidence` succeeded.
- Local release validation: `TEST_SHARDS=4 ./tests/release-check.sh --skip-docker` passed with 4,320 unit tests and 5,892 integration tests passing (53 expected integration skips). Docker coverage is supplied by the exact-SHA gated run above.

No candidate commit was added after this evidence was collected.

Issue #829, opened 2026-08-26
claude reader: subagent transcripts enumerate as phantom sessions (796 of 871) and are never harvested


itlackey:
## Summary

Two defects in `ClaudeCodeProvider`, both stemming from the reader having **no awareness of subagent transcripts**:

1. **Subagent transcripts enumerate as phantom top-level sessions** with fabricated ids. Measured on this machine: **796 of 871** listed sessions (91%) are subagent transcripts, not sessions.
2. **Subagent work is never attributed to its parent**, so it is never harvested into memories at all.

## Evidence

`ClaudeCodeProvider.listSessions({ sinceMs: 0 })`:

```
total sessions listed:                      871
listed as 'agent-<hash>' (subagent transcripts):  796
   {"sessionId":"agent-a057d00ca8257cc88","projectHint":"subagents","title":""}
   {"sessionId":"agent-a90f8db462c09a0c1","projectHint":"subagents","title":""}
   {"sessionId":"agent-aece043d5414d8206","projectHint":"subagents","title":""}
```

## Cause

`#walkJsonl(root)` delegates to `walkFiles` (`src/integrations/session-logs/provider-base.ts:44-54`), which **recurses into every subdirectory**. Subagent transcripts live one level down, so they are swept up:

```
~/.claude/projects/<project-slug>/<PARENT-SESSION-ID>/subagents/agent-<agentId>.jsonl
```

`listSessions()` then derives `sessionId` from the **filename** (`session-log.ts:125`), producing `agent-<agentId>` — an id that is not a session id and that nothing else in the system will ever ask for. `projectHint` becomes the literal string `"subagents"` rather than a project.

`grep -rn "subagent" src/integrations/harnesses/claude/` returns nothing: the reader has no concept of them.

## The linkage does exist

Worth correcting an earlier assumption (mine, in akm-plugins#101) that no on-disk artifact links a subagent transcript to a session. It does — you just link to the **parent session**, not to a hook id:

- The **directory name** is the parent session id: `<project>/<parent-session-id>/subagents/`.
- Every record inside the subagent transcript carries `sessionId` = the **parent session id**, plus its own `agentId`:
  ```json
  {"sessionId":"578b2d47-2382-4593-9423-950c7f4826f1","agentId":"a9104b2f0e5a70de9","type":"user","cwd":"..."}
  ```
[...]

PR #836, opened 2026-08-26
docs(extract): measure subagent-folding duplication and disclose re-extraction cost (#833)

### LINKED ISSUE #833: extract: measure whether folding subagent transcripts duplicates the parent's own summary (and disclose the re-extraction cost)
## Summary

Follow-up to #829 / PR #830, which folds a parent session's subagent transcripts into that parent's `readSession()`. Two consequences of that change were identified but deliberately not addressed there, and both deserve a measurement before the next release.

## 1. Content duplication

A parent transcript already contains each subagent's **final report** — that is the whole point of the Task tool's return value. PR #830 additionally folds in the subagent's **entire transcript**. So the same conclusions are now present twice: once as the agent's summary in the parent, once as the full working trace.

That may be fine (the trace carries reasoning the summary omits) or it may be actively harmful (the extractor sees a conclusion stated twice and over-weights it, or the budget pass drops parent content to make room for a subagent's tool-call noise). Nobody has measured which.

Worth checking: does folding change what gets extracted, and for the better? The honest test is extracting the same sessions with and without folding and comparing the candidates.

## 2. One-time re-extraction cost

`contentHash` is derived from `SessionData` (`extract.ts:565`), which now includes subagent events. Every previously-extracted session's hash therefore changes, so the next `--since` run re-extracts **all** of them — each substantially larger. Measured on one real session: **1583 -> 6738 events**, 5155 of them folded in from 32 subagent transcripts.

That is correct behavior — there genuinely is new content — but it is a one-time LLM-cost spike rather than a no-op, and it lands silently on whoever upgrades. The existing pre-filter caps per-event length and total chars, so no single call runs away, but the number of calls and their size both jump.

Worth deciding: is a note in the changelog enough, or should the upgrade path be staged?

## Possible outcomes

- measurement says folding improves extraction -> keep it, document the one-time cost
- measurement says it duplicates without adding -> fold only subagent events the parent does not already summarize, or drop folding and keep only #829's phantom-session fix
- measurement is ambiguous -> make folding the default but ensure the cost is disclosed on upgrade

## Context

- #829 — phantom sessions (796 of 871) and the missing subagent harvesting
- #830 — the fix; part 1 (phantom exclusion) is unambiguous and independent of this

itlackey:
Measurement follow-up to #829/#830, answering both questions #833 raised
before the next release: does folding a session's subagent transcripts
(#830) duplicate the parent's own summary, and what does re-extraction cost
on upgrade.

## Method

`readSession`, `preFilterSession`, and `buildExtractPrompt` are pure/
deterministic — no LLM call needed to see exactly what a real extraction
run would send to the model. For 3 real sessions on this machine I built two
fixtures each (the parent transcript alone, and the parent + its real
`subagents/*.jsonl` files copied verbatim) and ran the actual reader →
pre-filter → prompt-builder → `hashSessionContent` pipeline against both,
tagging every kept/dropped event by which file it came from (`filePath`) so
composition shifts are visible, not inferred.

| session | raw events (before → after folding) | pre-filter output composition (parent / subagent) | prompt chars | inline refs recovered |
| --- | --- | --- | --- | --- |
| `4f7f80f8…` (largest) | 1209 → 14224 | 204/0 → 204/0 (unchanged) | 92680 → 97528 | 2 → 162 |
| `578b2d47…` (the #829/#833 example) | 1583 → 6738 | 139/0 → 102/27 | 92048 → 93779 | 4 → 38 |
| `4a0d9e9b…` (smallest) | 155 → 2408 | 72/0 → 21/78 | 87535 → 88632 | 0 → 5 |

`578b2d47…`'s 1583 → 6738 raw-event jump is the exact figure #829/#833 cite,
confirming this is the same session originally measured.

## Q1 — does folding duplicate the parent's own summary?

Yes and no — it's a real tradeoff, not a clean win or loss:

- **Real benefit.** Inline `akm remember`/`akm feedback` calls a *subagent*
  made are recovered regardless of the 80,000-char pre-filter budget,
  because inline-ref extraction runs over the raw event stream, not the
  filtered one. On the largest session that's 162 refs recovered vs. 2
  without folding — this is literally #829's second defect ("subagent work
  is never harvested") getting fixed.
- **Real cost.** On sessions whose raw content is near or under the
  pre-filter's character budget, folding evicts a large share of the
  parent's own kept content to make room for subagent tool-call trace:
  parent-origin kept events dropped from 72 → 21 (**-71%**) on the smallest
  session and 139 → 102 (**-27%**) on the #829/#833 example. On the largest
  session the budget was already saturated by the parent's own tail alone,
  so folding changed nothing in the transcript section (all 204 kept events
  are parent-origin either way).
- **Duplication is measured, not hypothetical.** On the smallest session,
  subagent `agent-a4f581608db6e05b1` ("Verify V28+V32: dead surface")
  appears **twice** in the *same* 88KB prompt sent to the extraction LLM:
  once via its own folded final message, once via the parent's own
  `<task-notification>` record of that delegated call's result. A direct
  diff against a comparable agent's final message and its parent-side
  notification measured ~92% textual overlap (the parent's own record
  already nearly duplicates the subagent's own final text, independent of
  folding — folding just adds the second copy).

I considered narrowing the fold to drop a subagent transcript's terminal
event (its apparent "final report") to remove exactly this duplication
without losing the tool-call trace, and rejected it: #830's own regression
fixture (`tests/integration/session-logs-providers.test.ts`, "folds subagent
transcripts into the parent session (#829)") has a subagent transcript whose
**only** event is that terminal assistant turn — a single delegated
`akm remember` call. The same drop-the-last-event rule would strip the only
content in short single-step delegations, undoing the harvesting #830 added
and failing that exact test.

**Decision: keep folding as shipped.** Neither "remove it" nor "narrow it"
is clearly better than what's there, and the one narrowing considered would
cost more (breaking short-delegation harvesting) than it fixes.

## Q2 — one-time re-extraction cost

`contentHash` (`src/commands/improve/extract.ts:565`) hashes the full event
stream, which now includes folded subagent events, so every
previously-extracted session's hash changes the first time it's read on a
build with #830. The next `--since` run therefore re-extracts all of them
once — same per-session LLM-call shape as any first extraction, just larger
prompts (see the table above: +1.2% to +5.2% prompt chars in these 3 cases,
capped by the existing 80,000-char pre-filter budget so no single call runs
away). This is disclosed in the CHANGELOG entry below; it's a one-time cost
on upgrade, not a recurring one.

## What's shipped

- CHANGELOG.md — `[Unreleased]` entry with the measurement, the numbers, and
  the "keep folding" decision, for whoever hits the re-extraction spike on
  upgrade.
- No behavior change. #829's phantom-session exclusion and #830's folding
  both stay exactly as merged.

## Verification

No code changed, so no new regression tests; confirmed the existing suites
are unaffected:

- `npx tsc --noEmit -p tsconfig.json` — clean
- `bun run test:unit` — 3857 pass / 0 fail (matches baseline)
- `bun run test:integration` — 5598 pass / 52 skip / 0 fail (matches baseline)
- `npx biome check CHANGELOG.md` — not linted by biome (markdown is out of
  scope for the linter); the pre-commit hook's ~1300 warnings are pre-existing
  and untouched by this change.

Closes #833

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,50 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Documentation
+
+- **Measured whether subagent-transcript folding (#830) duplicates the
+  parent's own summary, and disclosed the one-time re-extraction cost
+  (#833).** Using the actual reader/pre-filter/prompt-builder code against 3
+  real sessions on this machine — no LLM calls; `contentHash`,
+  `preFilterSession`, and `buildExtractPrompt` are deterministic:
+  - Raw event counts grow 2x-12x once subagent transcripts are folded in
+    (measured: 1209 -> 14224; 1583 -> 6738, the exact session cited in
+    #829/#833's "1583 -> 6738" figure; 155 -> 2408). `contentHash` is
+    computed over that stream, so every previously-extracted session's hash
+    changes and the next `--since` run re-extracts all of them once, each
+    with a larger prompt (+1.2% to +5.2% prompt chars across the 3 sessions,
+    since the 80,000-char pre-filter budget caps how much of the growth
+    actually reaches the LLM).
+  - The result is a genuine tradeoff, not a clean win or loss. **Benefit:**
+    inline `akm remember`/`akm feedback` calls made *by subagents* are
+    recovered regardless of the budget cap (inline-ref extraction runs on
+    the raw event stream, not the pre-filtered one) — up to 162 refs
+    recovered on the largest session measured (was 2 without folding),
+    fixing #829's "delegated work is never harvested" defect. **Cost:** on
+    sessions whose raw content is near or under the pre-filter's character
+    budget, folding evicts a large share of the parent's own kept content to
+    make room for subagent tool-call trace — parent-origin kept events
+    dropped 27% and 71% respectively on the two smaller sessions measured.
+    On the largest session the budget was already saturated by the parent's
+    own tail, so folding changed nothing there. Duplication is real, not
+    hypothetical: on the smallest session, one subagent's conclusion appears
+    twice in the same prompt sent to the extraction LLM — once via its own
+    folded final message, once via the parent's own record of that
+    delegated call's result, which independently already captured ~92% of
+    the same text verbatim.
+  - A narrowing that drops a subagent transcript's terminal event (its
+    apparent "final report") to avoid this specific duplication was
+    considered and rejected: the existing #830 regression fixture has a
+    subagent transcript whose *only* event is that terminal turn (a single
+    delegated `akm remember` call) — the same rule would drop the only
+    content in short single-step delegations, undoing the harvesting #830
+    added.
+  - **Decision: keep folding as shipped.** The data does not cleanly favor
+    removing or narrowing it, and the one narrowing considered would cost
+    more than it fixes. #829's phantom-session exclusion is unaffected
+    either way.
+
 ## [0.9.2-alpha.3] - 2026-08-26
 
 ### Fixed


itlackey:
The measurement is solid and the "keep folding" call is defensible, but the framing offered two options — keep folding, or narrow it — when there is a third, and I think it is the right one to at least record.

**The budget is the binding constraint, and it is already parameterized.**

`preFilterSession`'s cap is `DEFAULT_MAX_TOTAL_CHARS = 80_000` (`src/integrations/session-logs/pre-filter.ts:42`). That number was sized for parent-only sessions. Folding multiplied raw input by roughly 9x on one measured session (1583 -> 6738 events; 155 -> 2408 on another) while the budget stayed fixed — so the 27% and 71% parent-content eviction is not an inherent property of folding. It is what happens when you multiply the input and hold the budget constant.

And `maxTotalChars` is not a constant that needs code to change. It is threaded end to end already:

- `PreFilterOptions.maxTotalChars` (`pre-filter.ts:116`, honored at `:175`)
- through `extract.ts` at `:570`, `:705`, `:1332`, `:1389`
- sourced from user config at `:1518` — `extractProcess?.maxTotalChars`
- and it is a published config field: `schemas/akm-config.json:1219`, `{"type":"integer","exclusiveMinimum":0}`

So "raise the budget for subagent-heavy sessions" is a **config change available today**, not a feature request.

**Why this matters for the decision.** The rejected narrowing was rejected for a good reason (#830's own fixture has a subagent whose only event is the terminal turn, so dropping terminal events would break exactly the harvesting #830 added). But the eviction finding is the more consequential half of this measurement and it survives that rejection: on budget-constrained sessions the extractor now sees **71% less of the parent's own content** than it did before #830. That is a real quality change to primary-session extraction, traded for delegated-work recovery.

Worth adding to the CHANGELOG entry alongside the re-extraction cost: operators who care about the parent's content on subagent-heavy sessions can set `maxTotalChars` above 80,000, and the measurement gives them the number to reason about. Something like:

> Sessions with many subagents may evict parent-origin content under the default 80,000-char pre-filter budget (measured: 27% and 71% on two real sessions). Raise `improve.strategies.*.processes.extract.maxTotalChars` if the parent's own content matters more than the token cost.

Also worth confirming before this is settled: is `hashSessionContent` genuinely stable across a `maxTotalChars` change? The doc comment at `extract.ts:487-489` asserts it hashes raw events precisely so config changes never move the hash. If that holds, raising the budget is free of a second re-extraction wave — which would make it a much easier recommendation.

Issue #828, opened 2026-08-26
extract: SessionEnd hook fails with "session <id> not found for harness claude-code" on 232 of 234 runs


itlackey:
## Summary

Session extraction is almost entirely non-functional via the Claude Code `SessionEnd` hook. **232 of 234** runs in `~/.local/state/akm-claude/extract.log` on one machine fail identically:

```json
{
  "schemaVersion": 1,
  "ok": false,
  "shape": "extract-result",
  "dryRun": false,
  "type": "claude-code",
  "sessionsProcessed": 0,
  "sessionsSkipped": 0,
  "candidatesCreated": 0,
  "proposals": [],
  "sessions": [],
  "warnings": [
    "session 889d142e-9f8c-4868-95ec-515a4ebc1c73 not found for harness claude-code"
  ],
  "durationMs": 53
}
```

`durationMs` in the 50-60ms range confirms it fails at lookup, before any session is read or any LLM is called.

## Why this matters

This is the actual reason durable memories are not being harvested — not #812. Only **2** of the 234 logged runs ever reached the JSON-parse path that #812 describes. Every akm-enabled session on this machine starts with:

> The last AKM session-end memory extraction failed — durable memories were not harvested.

and the cause is this lookup, not model output.

## Where

The warning is emitted around `src/commands/improve/extract.ts:1625`. The plugin invokes extraction from a `SessionEnd` hook with an explicit `--session-id`, and that id does not resolve against the claude-code harness's session enumeration.

## Diagnosis not yet done

Open questions worth splitting first — does a transcript file for one of the failing ids exist on disk at all?

- Claude Code transcripts live under `~/.claude/projects/<slugified-cwd>/<session-id>.jsonl`. If the harness enumerates only one project directory, or derives it from a cwd that differs from where the hook ran, every lookup misses.
- Timing: if `SessionEnd` fires before the transcript is flushed, the file may not exist yet at lookup time.
- Matching: exact id vs basename vs path-derived.

## Impact

Silent, total loss of session-derived memory capture, on the machine's steady-state path. `ok: false` and the populated top-level warning mean it is at least *visible* (that part works, courtesy of #812's fix) — but nothing consumes it, so it reads as a startup notice and nothing more.

itlackey:
Closing: this is not a CLI defect, and the fix already exists and is released — the affected machine is simply running a plugin build that predates it.

## Root cause

[...]

PR #838, opened 2026-08-26
health: report Claude harness plugin version, warn on stale/out-of-range

### LINKED ISSUE #832: health: report installed plugin versions and warn when a stale plugin is paired with the CLI
## Summary

`akm doctor` should report the version of the **installed harness plugin(s)** alongside the CLI's own version, and warn when the installed plugin predates the CLI it is paired with.

## Why — this cost a full investigation

#828 was filed as an akm CLI bug: session extraction failing on **232 of 234 runs** with `session <id> not found for harness claude-code`. Diagnosing it took a long, careful pass: enumerating transcripts, correlating session ids against `session-env/`, measuring session lifetimes from the plugin's rotated `events.jsonl`, and ruling out project scoping, write-after-`SessionEnd` timing, and subagent-transcript mismatch.

The answer: **not a CLI bug at all.** The fix already existed (akm-plugins#93), was merged, and had shipped in tag `v0.9.1202608250804`. The machine was simply running plugin `0.9.1`, marketplace clone pinned at `61b37e4` from three days before that merge.

Every piece of information needed to reach that conclusion in one step was sitting on disk:

- CLI version: `akm info` already reports it
- installed plugin version: `~/.claude/plugins/cache/akm-plugins/akm/<version>/.claude-plugin/plugin.json`
- marketplace clone HEAD: `git -C ~/.claude/plugins/marketplaces/akm-plugins rev-parse HEAD`

Nothing correlates them, so a stale plugin is indistinguishable from a broken CLI.

## What to add

In `akm doctor`, a check that:

1. reports each detected harness plugin and its installed version;
2. compares against the newest available (marketplace clone / published tag) and warns when behind, naming the update command;
3. ideally notes when the *installed* plugin's `AKM_VERSION_RANGE` does not admit the running CLI — the plugin already carries that contract, and a mismatch means the plugin has silently disabled itself.

Point 3 is the sharp one. The plugin's version gate is deliberately strict, and when it rejects the CLI both surfaces log it and degrade quietly (`version_out_of_range` / `akm_version_mismatch`). From the CLI side today there is no way to know that happened.

## Suggested output shape

```
plugins:
  akm@akm-plugins   installed 0.9.1   available 0.9.1202608250804   STALE
    -> claude plugin update akm@akm-plugins
    -> installed plugin requires akm-cli ^0.9.0; running 0.9.2-alpha.3 -> NOT ADMITTED (plugin is inactive)
```

## Context

- #828 — closed as not-a-CLI-bug; the full diagnosis is in the closing comment
- akm-plugins#93 — the fix that was already released but not installed

itlackey:
## Summary
- `akm health` now reports installed Claude Code `akm` harness plugin version(s) and warns when a plugin is behind the newest published tag (naming the update command) or when the plugin's own declared `AKM_VERSION_RANGE` no longer admits the running CLI (the plugin has silently disabled itself).
- Hooked into the existing `gatherAncillaryAdvisories` best-effort pipeline in `src/commands/health.ts`, next to `stash-git-exposure` / `binary-config-skew` — same shape, same degrade-to-benign-on-missing-input contract. New collector lives in `src/commands/health/plugin-staleness.ts`.
- Read-only: never installs, updates, fetches, or writes to the plugin cache or marketplace clone. The one network read (`git ls-remote --tags origin` against the marketplace clone's remote, 5s timeout, injectable seam) lists refs only — it's the one deliberate exception to "`akm health` makes no network call," because a local marketplace clone can't prove what's newest upstream (that's exactly how #828 happened).
- Added `AKM_CLAUDE_PLUGINS_DIR` env override (mirrors the existing `AKM_CLAUDE_PROJECTS_DIR` pattern) so the plugin cache/marketplace root resolves portably instead of hardcoding `~/.claude/plugins`, and wired it into `withIsolatedAkmStorage` so existing health tests stay hermetic.
- Added `satisfiesRange` to `src/registry/semver.ts` (thin wrapper over `semver.satisfies`).

## Real output on the (verified stale) dev machine
```
⚠ plugin-version  [warn, deterministic, high]  akm@akm-plugins: installed 0.9.1 — available 0.9.1202608250804 (STALE) — -> claude plugin update akm@akm-plugins — installed plugin requires akm-cli ^0.9.0; running 0.9.2-alpha.3 -> NOT ADMITTED (plugin is inactive)
```

Closes #832

## Test plan
- [x] `bun test tests/integration/commands/health/plugin-staleness.test.ts tests/integration/health-plugin-version-wiring.test.ts` — 13 pass / 0 fail
- [x] Mutation-tested the core detection logic (stale computation, range-admission check, malformed-manifest skip, malformed-range degrade, existsSync network guard, cache-dir-missing degrade) — each mutation broke ≥1 test; all reverted and green again
- [x] `bun run test:unit` — 3857 pass / 0 skip / 0 fail
- [x] `bun run test:integration` — 5611 pass / 52 skip / 0 fail (13 more than baseline ~5598, matching the new tests)
- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `npx biome check` on changed files — clean (repo-wide pre-existing warnings untouched)
- [x] Ran `akm health` for real against this machine's actual `~/.claude/plugins` — confirmed the stale/inactive plugin is surfaced as shown above

### DIFF
--- a/src/commands/health.ts
+++ b/src/commands/health.ts
@@ -18,6 +18,7 @@ import { readSemanticStatus } from "../indexer/search/semantic-status";
 import type { Database } from "../storage/database";
 import { closeDatabase, openReadonlyExistingDatabase } from "../storage/repositories/index-connection";
 import { queryTaskHistory } from "../storage/repositories/task-history-repository";
+import { pkgVersion } from "../version";
 import { collectImproveAdvisories } from "./health/advisories";
 import { HEALTH_CHECKS, type HealthCheckContext, runHealthEngineProbes } from "./health/checks";
 import {
@@ -35,6 +36,7 @@ import {
   computeEnrichmentMintingRollup,
   probeStateDbRoundTrip,
 } from "./health/metrics";
+import { collectPluginStalenessAdvisories } from "./health/plugin-staleness";
 import { collectStashExposureAdvisory, type GitRunner } from "./health/stash-exposure";
 import { collectSurfacesAdvisories, type EgressConfigView } from "./health/surfaces";
 import { buildPerRunSummaries } from "./health/task-runs";
@@ -361,11 +363,12 @@ function gatherImproveSummaryPhase(
 }
 
 /**
- * The three best-effort advisory groups beyond the health-check registry:
- * improve advisories, the `stash-git-exposure` probe, and the 08 surfaces
- * group (binary-config-skew, egress-endpoints). Order matches emission order in
- * the returned array. A probe/filesystem failure in either try/catch must not
- * abort the health report — each group degrades to "no advisory" independently.
+ * The four best-effort advisory groups beyond the health-check registry:
+ * improve advisories, the `stash-git-exposure` probe, the 08 surfaces group
+ * (binary-config-skew, egress-endpoints), and `plugin-version` (itlackey/akm#832).
+ * Order matches emission order in the returned array. A probe/filesystem
+ * failure in any try/catch must not abort the health report — each group
+ * degrades to "no advisory" independently.
  */
 function gatherAncillaryAdvisories(
   db: Database,
@@ -408,6 +411,17 @@ function gatherAncillaryAdvisories(
         config: egressConfigView,
       }),
     );
+  } catch {
+    // Non-fatal.
+  }
+
+  // itlackey/akm#832: report installed Claude Code harness plugin version(s)
+  // and warn when stale or when the plugin's own akm-cli version range no
+  // longer admits this CLI. Best-effort — no plugin installed, an unreadable
+  // manifest, or a network failure while checking the newest tag must not
+  // abort the health report.
+  try {
+    advisories.push(...collectPluginStalenessAdvisories({ cliVersion: pkgVersion }));
   } catch {
     // Non-fatal.
   }
--- a/src/commands/health/plugin-staleness.ts
+++ b/src/commands/health/plugin-staleness.ts
@@ -1,0 +1,254 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+/**
+ * `plugin-version` advisory for `akm health` (itlackey/akm#832).
+ *
+ * #828 was filed as a CLI bug — session extraction failing on 232/234 runs —
+ * and took a full investigation to resolve. The actual cause: the harness
+ * plugin (`akm@akm-plugins`, installed in Claude Code's plugin cache) was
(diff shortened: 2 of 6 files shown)


itlackey:
Verified the network-call question rather than waving it through, because you flagged it yourself and it is the one real design decision in this PR.

**You are right that the offline path cannot catch #828.** I checked this machine:

```
installed plugin version:  0.9.1
marketplace clone HEAD:    61b37e4 chore: bump version to 0.9.1
marketplace clone tags:    v0.9.1
```

The local clone is at the *same* commit as the installed plugin, and does not have the `v0.9.1202608250804` tag at all. So a purely local comparison would have reported "up to date" — the exact false negative that made #828 take a full investigation. `git ls-remote` is genuinely what closes that gap.

**On the convention you deviated from.** It is implicit, not documented — I could not find an offline guarantee for `health` anywhere (only `html-report.ts:28` documents the reverse, that its report *requires* network). And `stash-exposure.ts:23` already shells out to `git` with a 5s timeout, so subprocess-git in an advisory is established. The distinction is narrower than "no network": existing advisories touch **local** git, and `egress-endpoints` deliberately *lists* remote destinations without contacting them (`surfaces.ts:11`). `ls-remote` is the first advisory that actually reaches out.

**The part worth deciding explicitly.** The check has two independent halves with different cost profiles:

| half | needs network | would it have caught #828? |
| --- | --- | --- |
| range-admission mismatch (plugin has silently disabled itself) | **no** | catches the *inactive plugin* symptom |
| staleness vs upstream | **yes** | catches the *behind upstream* cause |

Both are valuable and they catch different things. That suggests an option neither of us raised: keep the range-admission half unconditional (free, offline, high signal) and let the upstream probe be the part that can be skipped — via `--offline`, an env var, or simply being skipped when the existing egress/offline config says so. That preserves the whole value of the check for anyone who wants it while not making `akm health` reach the network by default in an air-gapped or privacy-sensitive setup.

I am not asking for a rewrite — the 5s timeout, the injectable seam, fetch-nothing-just-list-refs, and degrade-to-no-claim are all the right mitigations, and the inline documentation is honest. This is a call for the repo owner: **is `akm health` allowed to make an outbound request by default?** If yes, this ships as-is. If no, the split above keeps most of the value offline.

Worth also confirming: does the existing egress/endpoint configuration already express a user preference this probe should honor? If there is a configured way to say "do not contact remotes," this should respect it rather than introduce a second, separate control.

itlackey:
Owner decision on the network question: **health may make the network call when the network is available, and must degrade gracefully when it fails or is unavailable.** No offline gating required.

The implementation as shipped already matches this — 5s timeout on `git ls-remote`, any failure degrades to "no staleness claim" rather than an error or a false warning, and the offline half (range-admission) works regardless. No changes needed on this point.

Issue #840, opened 2026-08-26
extract: determine fold-vs-link for subagent sessions; explore chunked extraction instead of truncation


itlackey:
## Summary

Owner-directed design determination, two questions:

**Q1 — fold vs link.** #830 folds each subagent's full transcript into the parent's event stream before extraction. The alternative: do NOT fold; extract the parent alone, and represent delegated work as **links** in the extracted output (the parent's task-notification already summarizes each subagent; provenance metadata — `agentType`, `description`, parent session id — is all on disk). Determine which is better, with measurements, and implement the winner.

What folding buys (measured, #836): inline `akm remember`/`akm feedback` calls made *by subagents* are recovered — 162 refs vs 2 on the largest session — because inline-ref harvesting runs over the raw stream before the budget cap. What folding costs (measured): 27–71% of parent-origin content evicted under the fixed 80k pre-filter budget, plus the doubled-conclusion duplication (#839).

Note the decomposition that reframes Q1: inline-ref recovery does NOT require folding into the *prompt* — refs are harvested pre-filter. A "link, don't fold" design could keep raw-stream ref harvesting over subagent transcripts while sending only the parent's prose to the LLM. Evaluate that hybrid as a first-class option, not just the two extremes.

**Q2 — chunking instead of a truncation budget.** The current pre-filter is a fixed `maxTotalChars` (default 80,000) with recency-biased eviction — "keep the newest, drop the oldest" — a heuristic designed for a single agent's narrative and now demonstrably misaligned for merged multi-agent streams (the newest content is often subagent tool noise). Explore replacing truncation with **chunked extraction**: split an over-budget session into coherent chunks (per-subagent is a natural boundary; time-windows for the parent), run extraction per chunk, and merge candidates. Map-reduce, not window-slide.

Chunking interacts with Q1: per-subagent chunks make "fold" and "link" nearly converge — each subagent is extracted in its own call and attributed to the parent. The determination should treat "chunked extraction with per-subagent chunks" as a candidate architecture that may dominate both Q1 options.

Costs to quantify honestly: chunking multiplies LLM calls per session (bound it: chunks per session on real data at 80k chunk size); candidate merging needs a dedupe rule; `contentHash` idempotency (hash raw events, `extract.ts:487-489`) must survive whatever is chosen — config changes must never move the hash.

## Deliverable

[...]

Issue #839, opened 2026-08-26
extract: dedupe the subagent final report against the parent's task-notification copy; pin no-double-extraction


itlackey:
## Summary

Two follow-ups to #830's subagent folding, both owner-directed:

**1. Dedupe the doubled subagent conclusion.** Each subagent's final report now appears **twice** in the same extraction prompt: once as its folded final message, once as the parent's `<task-notification>` record of that same call. Measured ~92% textual overlap on one real pair (PR #836's measurement, session `4a0d9e9b…`, subagent `agent-a4f581608db6e05b1`). Duplicated content burns pre-filter budget while parent-origin content is being evicted (27–71% measured eviction under the 80k cap).

**Direction that avoids the known trap:** drop the **parent's task-notification copy**, keep the **subagent's original**. The naive inverse (drop the subagent's terminal event) was evaluated and rejected in #836 — some subagent transcripts consist *only* of that terminal turn (e.g. a single delegated `akm remember`), so dropping it destroys exactly the harvesting #830 added. Deduping from the parent side has no such failure mode: the notification is by definition redundant with the original it summarizes.

Care point: the parent's notification carries metadata the subagent transcript may lack in-stream (the notification frames *why* the parent delegated). Dedupe should remove the redundant **body**, not necessarily the whole event — evaluate whether to keep a one-line stub (`[subagent <id> completed: <description>]`) so the parent's narrative stays coherent.

**2. Pin the no-double-extraction guarantee with a regression test.** Verified today: both discovery and `--session-id` lookup route through `listSessions()` (`extract.ts:1608-1611`), which excludes `subagents/` dirs since #830 — so a subagent cannot currently be extracted as its own session *and* folded into its parent. But nothing tests that property end to end. Add a regression test: given a parent with subagent transcripts, (a) discovery-mode extraction processes exactly one session, (b) `--session-id agent-<hash>` does NOT resolve, (c) the folded content is attributed only to the parent's contentHash. Without this pin, a future enumeration change silently reintroduces double-extraction.

## Relationship to #840

If the fold-vs-link determination (#840) concludes against folding, part 1 becomes moot and part 2 still stands.

itlackey:
Already implemented and merged: #842 (fix(extract): dedupe the doubled subagent conclusion, pin no-double-extraction), merged into `release/0.9.2`. Assigned to the 0.9.2 milestone; this issue will close automatically once `release/0.9.2` merges to `main` via #822.

PR #841, opened 2026-08-26
docs(extract): determine fold-vs-link for subagent sessions (#840)

itlackey:
## Summary

Owner-directed design determination for #840, both questions:

- **Q1 (fold vs link):** measures all three prompt-shaping candidates plus
  the hybrid the issue calls out as a first-class option — fold+dedupe (as
  shipped by #830, plus #839's proposed dedupe), link-only, and a
  "harvest-without-prompting" hybrid (raw-stream inline-ref harvesting stays,
  the LLM prompt is parent-only).
- **Q2 (chunking):** measures chunked map-reduce extraction (per-subagent
  chunks + parent time-windows) against the same real sessions, bounding the
  LLM-call multiplier on real data as the issue asks.

**Recommendation: the harvest-without-prompting hybrid.** On every measured
session it matches #830's inline-ref recovery exactly (162/162, 38/38, 1/1,
8/8) with zero parent-content eviction (vs 27.5%/28.6% evicted under fold on
two of four sessions), and on the largest multi-subagent session its prompt
is smaller than fold's while recovering the same refs — strictly dominating
fold on every measured axis there. Chunked extraction was measured at
**9x-229x more LLM calls per session** on real data and is not recommended.

**#839's dedupe was measured to have zero effect on the actual LLM prompt**
on all four sessions available: the duplicate `<task-notification>` content
it targets is already evicted by the recency-biased 80k budget before dedupe
would matter (89 raw duplicate pairs found across four sessions, 0 with both
copies surviving into the pre-dedupe prompt). This doesn't mean the defect
is imaginary — #836's own worked example is real, just on a session whose
parent transcript is no longer on this machine to re-measure — but it does
mean the practical cost of folding is eviction, not duplication, which is
exactly what the recommended hybrid fixes and dedupe does not.

The doc also flags a real coordination hazard for whoever implements #839:
its "drop the parent's notification, keep the subagent's own copy" direction
inverts under the hybrid winner, where the parent's notification is the
*only* surviving copy in the prompt — applying #839 unconditionally under
the hybrid would delete the sole remaining trace of delegated work. See
§6 of the doc.

## What's in this PR

- `docs/plans/subagent-extraction-design.md` — the decision document: method,
  measured tables across all four candidates on four real sessions (#836's
  three plus one added for chunking-cost coverage, since one of #836's three
  no longer exists on disk), the two invariants any winner must preserve
  (`hashSessionContent` idempotency, no-double-extraction) with code
  references, the `maxTotalChars` config-knob disposition, the determination
  argued from the numbers, and a scoped implementation plan for the winner
  (not implemented here). The reproducible measurement script is embedded in
  §7.
- `CHANGELOG.md` — `[Unreleased]` entry summarizing the determination.

**No behavior changes.** `src/` is untouched — #839 is in flight in the same
files and the winning design builds on it (see the coordination note above),
so this PR is doc + plan only, per the issue's own scope guidance.

## Verification

- `npx tsc --noEmit -p tsconfig.json` — clean
- `bun run test:unit` — 3857 pass / 0 fail (matches baseline)
- `bun run test:integration` — 5598 pass / 52 skip / 0 fail (matches baseline)
- `bun run lint` — all gate scripts OK, including `lint-doc-examples` and
  `lint-active-docs-terminology` against the new doc; the ~1300 biome
  warnings are the same pre-existing baseline noted in #836, untouched here

Ref #840. Not closing it — the determination and plan are complete, but the
winning design isn't implemented yet (deliberately, per scope).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Documentation
+
+- Recorded the fold-vs-link subagent-extraction design determination in
+  `docs/plans/subagent-extraction-design.md` (#840). Measured four candidates
+  (fold+dedupe as shipped, link-only, a harvest-without-prompting hybrid, and
+  chunked map-reduce extraction) on the same real sessions #836 used plus one
+  added for scale. Headline: the hybrid recovers 100% of #830's inline-ref
+  harvesting (162/162, 38/38, 1/1, 8/8 across the four sessions) with zero
+  parent-content eviction (vs 27.5%/28.6% evicted under fold on two of the
+  four), and #839's dedupe was measured to have zero effect on the actual
+  LLM prompt on all four sessions (the flagged duplicate content is already
+  evicted by the recency-biased budget before dedupe would matter). Chunked
+  extraction was measured at 9x-229x more LLM calls per session on real
+  data and is not recommended. No behavior changes shipped in this PR.
+
 ## [0.9.2-alpha.3] - 2026-08-26
 
 ### Fixed
--- a/docs/plans/subagent-extraction-design.md
+++ b/docs/plans/subagent-extraction-design.md
@@ -1,0 +1,676 @@
+# Subagent extraction design determination (#840)
+
+**Status:** decided — recommend Candidate 3 ("harvest-without-prompting
+hybrid"); scoped implementation plan below, not yet implemented.
+**Companions:** #830 (fold, shipped), #836 (the measurement this reuses and
+extends), #839 (dedupe + no-double-extraction pin, in flight — assumed to
+land; not implemented here).
+
+This is a decision document, not an implementation. No `src/` files change in
+this PR. All numbers below come from running the real, unmodified
+`ClaudeCodeProvider.readSession` → `preFilterSession` → `buildExtractPrompt`
+pipeline against real session logs on this machine — no LLM calls, fully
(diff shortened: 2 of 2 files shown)


itlackey:
Independently reproduced the headline measurements by extracting §7's script from the doc verbatim and pointing it at two of the sessions (the script ships with a placeholder path by design, since sessions are machine-specific — filled in locally).

**Reproduced exactly:**
- `578b2d47…`: link prompt **92,044** / fold prompt **93,781** chars — matches the doc's table to the digit; 27.5% parent eviction — matches; 38 inline refs — matches.
- **fold+dedupe prompt == fold pre-dedupe prompt** on both sessions (93,781 == 93,781; 92,586 == 92,586) — the zero-prompt-effect finding for #839 reproduces. Raw duplicate pairs found: 3 and 55 on these two sessions, consistent with the doc's 89-across-4.
- hybrid == link prompt chars on `393273fd…` (91,781) while keeping fold's ref recovery — the winner's central claim holds.
- chunking on `393273fd…`: **229 chunks → 229 LLM calls vs 1 today** — matches, and confirms the disqualifying cost for the background pass.

**One drift, worth a note in the doc:** `393273fd…` measured 24.2% parent-char eviction here vs 28.6% in the table (events 16.1% vs the doc's figure). That session appears to still be live and has grown since the agent measured — expected for a moving corpus, and it does not change any conclusion. A one-line caveat that the table is a point-in-time snapshot of mutable sessions would preempt the question.

The #839 coordination hazard (under this design the parent's notification is the only surviving copy, so an unconditional dedupe would delete it) has been relayed to the in-flight #839 implementation with the required scoping.

PR #848, opened 2026-08-29
docs: add manual test suite for task/workflow (0.9.2-alpha.5)

itlackey:
## Summary

- Adds `docs/architecture/testing/task-workflow-manual-tests.md`, a single reviewed manual QA suite for `akm task` and `akm workflow` (0.9.2-alpha.5, commit `fb7bef3f`), assembled from three parallel draft agent suites (task-only, workflow-only, task↔workflow seam) after adversarial review.
- Review pass: spot-checked ~11 exact JSON fields/error codes/exit codes live via `bun src/cli.ts` in isolated sandboxes; all matched except one, which is corrected in the doc (a claim that `akm workflow --help` omits the `plan` subcommand does not reproduce on this commit). Resolved one previously-`UNVERIFIED` marker live (the parent-side child-run evidence shape for a `uses: tasks/...` dispatch).
- Folds in **akm#846** (`task sync` orphan-removal is scoped by bundle *name*, not path — a name collision can make an isolated test bundle remove another bundle's real scheduler entries) and **akm#847** (a `blocked` run that's abandoned can never be resumed, contradicting `abandon --help`) as known, already-filed issues rather than re-discovering/re-filing them; the corresponding tests assert the current buggy behavior and are explicitly labeled expected-to-fail pending those issues.
- 38 tests total: 16 non-destructive task tests, 10 workflow tests, 5 non-destructive integration/seam tests, plus 7 destructive/platform-gated tests (task-add/sync scheduler mutation, cross-bundle collision reproduction, and macOS/Windows platform-gated placeholders) in a clearly separated final section with mandatory crontab backup/restore steps.
- Notes that the AKM stash asset `knowledge/projects/akm/akm-manual-testing-checklist` (dated 2026-05-17) has stale Tasks/Workflows sections referencing nonexistent commands (`akm tasks list`, `akm workflow next`) and recommends replacing those sections with a pointer to this document — does not edit that asset itself.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] Docs-only change; `bun run test:unit` / `test:integration` untouched
- [x] Did **not** run `akm task sync`/`akm task add` against the real scheduler; verified the host crontab is unchanged before/after this work

### DIFF
--- a/docs/architecture/testing/task-workflow-manual-tests.md
+++ b/docs/architecture/testing/task-workflow-manual-tests.md
@@ -1,0 +1,1814 @@
+# Manual tests — `akm task` and `akm workflow` (0.9.2-alpha.5)
+
+This document is a focused manual-QA suite for the `akm task` and `akm
+workflow` subsystems shipped in PR #844 ("Task/workflow final refactor:
+typed task inputs, child workflows, v3 retirement"). It sits alongside
+[`manual-testing-checklist.md`](./manual-testing-checklist.md) (the
+general-purpose runbook, whose §14/§15 give a lighter pass over the same
+commands) as the deeper, field-level reference for this specific feature
+area — use it when validating a task/workflow release or investigating a
+regression in scheduling, typed inputs, child workflows, or the
+task↔workflow seam.
+
+**Verified against:** `akm` `0.9.2-alpha.5`, commit `fb7bef3f` (branch
+`release/0.9.2`). Every expected result marked "verified live" below was
+captured by actually running that exact command against that exact commit
+during authoring/review of this document, via `bun src/cli.ts` in an
+isolated sandbox (see [Setup](#setup-one-sandbox-for-every-non-destructive-test)
+below) — not inferred from source or docs. Anything not run live is marked
+`UNVERIFIED` in bold, inline, at the point it applies. **Re-verify the exact
+JSON field names, error strings, and exit codes here before relying on them
+against a later release** — none of this is covered by automated tests
+(that's the point of a manual suite), so it can drift silently.
+
+## Read this first — the one safety-critical finding
+
+**`akm task sync` (and, to a lesser extent, `akm task add`) write to the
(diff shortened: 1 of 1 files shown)


itlackey:
Heads-up for this suite's `task sync` scenarios — the hazard they were written around is being fixed.

**#846 has been narrowed** to just the correctness fix: scope the reconcile by resolved bundle **path/id** instead of display name. That is the change that closes the cross-bundle deletion risk, and it is small enough to land in 0.9.2 (milestone assigned).

**The dry-run/preview mode split out to #849** (milestone 0.10.0). That was the other half of the original issue, and it is the part that would let a tester *preview* a reconcile rather than run it live.

Implementation of #846 is in flight; PR to follow against `release/0.9.2`.

**What this means for the tests here once #846 lands:**

- **DT-5** (the cross-bundle collision reproduction) inverts. Today it documents the hazard — two bundles sharing a directory basename, where A's `sync` reaches for B's bindings. After the fix it should assert the opposite: A's `sync` leaves B's bindings alone, and a removal it cannot establish ownership for is refused with a clear error rather than fingerprint-guard luck. Worth keeping the test and flipping the expectation, since it is exactly the regression guard for this bug.
- **DT-3 / DT-4** (sync drift, shell-target and workflow-target) keep their mandatory `crontab -l > backup.txt` precondition and restore step regardless. The scoping fix removes the cross-bundle blast radius, but `sync` still mutates the real scheduler and still has no preview until #849 ships — so the DO-NOT-RUN-ON-A-MACHINE-WITH-REAL-SCHEDULES framing stays correct.
- The `UNVERIFIED` markers on those scenarios stay until someone can execute them safely; #849 is what makes that practical.

No change requested to this PR — the suite is accurate as written for the current code, and it is the artifact that surfaced #846 and #847 in the first place. This comment is so whoever merges the #846 fix knows DT-5 needs its expectation flipped in a follow-up.

Issue #891, opened 2026-08-31
worktree-concurrency suite is green only on an idle machine, and its non-blocking coverage is now gone


itlackey:
Follow-up from the 0.9.8 stabilization sprint. Two related problems in `tests/integration/workflows/worktree-concurrency.test.ts`.

## 1. Four tests pass in isolation and fail under external git load

Observed three times today, independently: by me on `main`, by the storage agent (which checked with and without its changes stashed), and during merge verification. Five failures in a full 8-shard `bun run test:integration`, **16/16 pass when the file runs alone**:

- `createUnitWorktree — concurrent units on one base repo > a fan-out of units minting worktrees at once all succeed with distinct, valid checkouts`
- `… > interleaved concurrent creates and cleanups leave no stale or missing registrations`
- `cleanupUnitWorktree — run root lifecycle > the run root is removed once its last unit worktree is cleaned up`
- `… > a retained dirty worktree keeps its run root alive`

Representative failure: `expect(cleanup.removed).toBe(true)` received `false`.

**Cause.** Each test already runs against its own private temp repo, so this is not fixture bleed. `src/workflows/exec/worktree.ts` serializes `git worktree add|prune|remove` against a base repo's administrative state through a **process-global** — and that global cannot see git activity from *other processes*. When the machine has concurrent git load (in this case several agents working in git worktrees of this repo, which had 7 registered at the time), the assertions fail.

These four assert real behavior — race-free worktree bookkeeping — so they should NOT be deleted. But a suite that is green only on an idle machine is a suite people learn to re-run rather than trust, which is how the two flakes deleted in `722117b6` survived as long as they did.

Options, none yet validated:
- make the tests tolerant of external contention without weakening the assertion (hard to do without masking a real regression — explicitly rejected during the sprint for that reason);
- run them against a base repo location no other process can plausibly touch, and confirm that is actually what fails today rather than assuming;
- accept the fragility and quarantine them out of the default parallel run, so a failure means something.

The one thing not acceptable is retry-around-the-assertion, which would hide a genuine regression.

## 2. Deleted coverage with no replacement

`worktree git calls do not block the event loop` was deleted in this sprint. It counted `setInterval` ticks at 5ms during real git calls — a scheduling assertion of the same class as the two flakes removed in `722117b6`, so the *method* was unfit.

But **the property it asserted has no other test in the suite**: that git calls are genuinely async and do not freeze the event loop. That is a real gap, deliberately recorded rather than left silent.

[...]

Issue #889, opened 2026-08-31
Clean up $STASH/.akm: 82% is dead pre-0.9.0 residue, and nothing states what may live there


itlackey:
`$STASH/.akm` on a real stash is **165 MB** and carries **1,523 committed files**. Most of it is not current state.

akm already has a correct, documented XDG layout (`docs/architecture/internals/storage-locations.md`): `$CONFIG` `~/.config/akm`, `$CACHE` `~/.cache/akm`, `$DATA` `~/.local/share/akm`, `$STATE` `~/.local/state/akm`, `$STASH` `~/akm`. The project has also been actively migrating out of `.akm` — proposals → `state.db` (0.9.0), registry-index → `index.db` (0.8.0), `events.jsonl` → `state.db`, config-backups consolidated into `$CACHE`.

**The migrations happened. The files were never cleaned up.** This issue is the cleanup plus the missing rule.

> **Scope note.** The live writers that belong in `$STATE`/`$CACHE` (`distill-rejected`, `eval-cases`, `measurement`, `unresolved-sources`, the lock files) are split into **#890** for 0.10.0, because moving them changes user-visible paths and needs a migration note. Everything below is patch-safe.

## Tier 1 — dead residue: nothing in `src/` writes it

Verified by grep across `src/`; each has zero live writer.

| Path | Size | Status |
|---|---|---|
| `.akm/proposals/` | **135 MB** | Superseded by the `proposals` table in `$DATA/state.db` (0.9.0). Referenced in `src/` only by a comment in `core/state/migrations.ts:107`. |
| `.akm/runs.archived-2026-05-24T…/` | 13 MB | Orphaned archive of a directory that no longer exists |
| `.akm/archive/` | 11 MB | Legacy. Current prune writes `.akm/memory-cleanup/archive/` — a different path |
| `.akm/graph.json` | 607 KB | Superseded by `graph_*` tables in `index.db`; no writer |
| `.akm/consolidate-journal.json` | — | `storage-locations.md` already marks it "Safe to remove" |
| `.akm/proposals.db` | 0 bytes | Empty |
| `.akm/mv-transactions/` | — | Only a comment in `core/fs-txn.ts:16`, about collapsing legacy paths |

**135 MB of 165 MB (82%) is the dead proposals directory alone.**

[...]

Issue #888, opened 2026-08-31
Decide noNonNullAssertion: adopt it or turn it off — 1,374 inherited warnings make `bun run lint` output unreadable


itlackey:
`bun run lint` emits **1,410 warnings**, and **1,374 of them (97%) are one rule**: `lint/style/noNonNullAssertion`. Biome truncates the output ("Diagnostics not shown: 1354") and tells you to raise `--max-diagnostics`.

## The rule was never chosen

`biome.json` enables the recommended set and overrides exactly one rule:

```json
"linter": {
  "enabled": true,
  "rules": {
    "recommended": true,
    "suspicious": { "noExplicitAny": "error" }
  }
}
```

`noNonNullAssertion` is inherited from `recommended`, where it defaults to `warn`. Nobody decided to adopt it; it arrived with the preset and the codebase has been writing `!` ever since. The result is a rule that is on, ignored, and unenforced.

## Why this is worth closing out

A 1,374-entry advisory channel is not a signal, it is a wall. Concretely, today:

- `bun run lint` failed with `Found 1 error. Found 1410 warnings.` The error was a **single formatting diff in one test file**, and finding it meant per-file bisection because the real diagnostic was buried under warnings and then truncated.
- The pre-commit hook prints the same wall on every commit, which trains everyone to scroll past the one place a genuine error would appear.

This is the same failure mode as #886 (a workflow advisory firing on 100% of steps): an always-on warning nobody can act on devalues the channel that real findings share. Worth fixing for the same reason, and this one is 13x larger.

## Ask

Pick one. Both are defensible; the current state is not.

**A. Turn it off.** `"style": { "noNonNullAssertion": "off" }`. Honest about what the codebase does — `arr[i]!` after an explicit bounds check is idiomatic here and appears throughout hot loops (`base-linter.ts`, `expressions.ts`, `chunking.ts`). Cost: none. Benefit: `bun run lint` output becomes readable immediately.

**B. Adopt it.** Set it to `error` and fix the 1,374 sites. Biome marks a large share `FIXABLE` via optional chaining (`arr[i]!` → `arr[i]?`), but that is **not a safe blanket rewrite** — Biome itself labels it "Unsafe fix", and it changes semantics: an assertion that would have thrown becomes a silent `undefined` flowing downstream. Each site needs a judgment call, which is why this is real work rather than a codemod.

[...]

Issue #887, opened 2026-08-31
lint-doc-examples scans gitignored paths, so `bun run lint` fails locally but passes in CI


itlackey:
`bun run lint` currently fails on a clean checkout of `main` for anyone whose working tree contains gitignored draft docs. CI passes, because CI never has those files.

## Repro

With any `.md` file under `docs/.pending/` (gitignored — `.gitignore:34`, `docs/.pending/`):

```
$ bun run lint
docs/.pending/akm-artifact-provenance-workflow-improvements-brief.md
  line 135: unknown subcommand `artifacts` under `akm workflow`
  line 374: unknown flag `--out` for `akm workflow`
  line 571: unknown subcommand `show` under `akm workflow`
  line 595: unknown flag `--negative` for `akm workflow`
  ...
error: script "lint" exited with code 1
```

Two such files exist in my tree; both are unpublished drafts that reference `akm workflow` subcommands that were never built.

## Cause

`scripts/lint-doc-examples.ts` walks `DOC_ROOTS = ["docs", "src/assets/hints"]` (line 71) with a plain recursive `readdirSync` (`walkMarkdown`, line 359) that has no gitignore awareness and no dot-directory skip:

```ts
for (const entry of entries) {
  const full = path.join(dir, entry.name);
  if (entry.isDirectory()) walkMarkdown(full, out);
  else if (entry.isFile() && entry.name.endsWith(".md")) out.push(full);
}
```

So `docs/.pending/` is scanned exactly like published docs.

## Why it matters

The linter's premise is sound: docs must not teach commands that do not exist. But it should only enforce that on docs the repo actually ships. A gitignored draft is by definition not shipped, and holding it to the published-docs contract:

- breaks the local gate while CI stays green — the worst split, because the failure is invisible to everyone who could fix it and unavoidable for the one person who cannot;
- makes `bun run lint` unusable as a pre-commit check for anyone who drafts in-tree;
- cannot be fixed by fixing the drafts, since the whole point of `.pending/` is that they are unfinished.

I hit this while verifying an unrelated change and had to confirm by hand that the failure predated my work, which is exactly the wasted motion a broken gate causes.

## Ask

Make the scan ignore what git ignores. Options, cheapest first:

1. Skip dot-directories in `walkMarkdown` — one line, fixes `.pending/` and any future `.`-prefixed scratch dir.
2. Filter the collected file list through `git check-ignore --stdin` and drop matches. Exact, one subprocess, degrades gracefully outside a git checkout.
[...]

PR #893, opened 2026-09-01
0.9.8: cleanup and stabilization sprint (−20,099 net lines)

itlackey:
55 commits since `v0.9.7`. **490 files, +4,930 / −25,029 — net −20,099 lines.**

A cleanup and stabilization sprint. The bulk is deletion of machinery that policed the codebase's *shape* rather than its behavior.

## Closed issues

| # | what |
|---|---|
| #885 | `writeContradictEdge` was dead code while a drifted private copy did the real writing |
| #886 | deleted the no-output-schema workflow advisory |
| #887 | doc linter scanned gitignored paths; line-pinned waivers replaced with a content rule |
| #888 | `noNonNullAssertion` turned off |
| #889 | `$STASH/.akm` dead-residue advisory + opt-in cleanup + the missing storage rule |
| #891 | worktree suite green only on an idle machine |
| #768 | GitHub Actions pinned to commit SHAs |
| #770 | assessed and closed — most of it was already satisfied |
| #785 | isolation grandfather list drained 58 → 2 |
| #786 | unit/integration classification rule applied across ~250 files |
| #787 | 31 status-only assertions pinned to real observed values |
| #788 | near-duplicate clusters drained |
| #789 | real-server timing suite de-flaked |

## Two product bugs, both found while chasing what looked like test problems

**`GIT_TIMEOUT_MS` was a flat 30s with no margin (#891).** Reproduced deterministically with a slow-`git` shim: a healthy `git worktree remove` on a busy machine returned `{ removed: false, error: "timed out after 30000ms" }`. Anyone running a workflow with `isolation: worktree` alongside other git activity could hit spurious create failures or have a clean worktree wrongly retained. The issue's own theory — a process-global lock racing across processes — was **refuted** by load testing: 20–65× slower under contention, never once a wrong result.

**`MAX_LEXICAL_QUERY_TOKENS = 16` silently truncated search queries.** Past 16 unique tokens the tail was dropped — for natural-language input, usually where the discriminating words are. It also fed ranking, so token-overlap scoring ran on the truncated set too. Unexplained in code and in the commit that introduced it, unreachable from any flag. Measured before removing the bound: the planner handles 10,000 tokens in 9ms, so nothing was being protected.

## What was deleted

- **14 architecture ratchets** (function-size, import-cycle, seam guards) — fired on refactors that harmed nothing
- **15 golden/snapshot suites** + fixtures — "fixed" by regenerating, which proves nothing
- **12 characterization files** — pinned "what the code does today" by definition
- **8 of 11 lint scripts** (−3,666 lines). `lint-tests-isolation` alone was 717 lines standing in front of `src/core/paths.ts`, which *already* throws `TEST_ISOLATION_MISSING` at runtime
- **Both `MIN_TESTS` floors** — arbitrary numbers whose only job was to fight test deletion
- Silent truncation caps: `MAX_CONTENT_CHARS` ×8, `MAX_BODY_CHARS`, and the `MAX_URLS` scaffolding

## What was deliberately kept

- **`registry-network-boundary.test.ts`** — removed by the ratchet sweep, restored on review. Unlike the other 13 it is not a shape ratchet: it enforces that only `pinned-transport.ts` may import `node:http`/`node:https`, which is what stops a raw request bypassing address-pinned fetching. No surviving lint script covers that boundary.
- **3 lint scripts** that catch user-visible problems: secrets must route through the resolver, no dead refs in shipped assets, no docs teaching commands that do not exist.
- **Contract/envelope and spec-conformance tests.** Two deletion passes came back nearly empty and both were verified rather than accepted: all 54 CLI-envelope tests have unique names asserting different commands end-to-end, and the robots.txt suite is numbered spec conformance (BOM stripping, CRLF, group precedence). Cutting those removes real coverage, not sediment.

## Suite

| | before | after |
|---|---|---|
| unit | 4,401 | 6,326 (29s) |
| integration | 5,855 | 3,421 (66s) |
| total | 10,256 | 9,747 |

Unit grew only because #786 moved ~230 files out of integration; the totals fell.

Verified on the merged tree: tsc clean, lint clean, both suites green on consecutive runs.

## Note

`package.json` is still `0.9.7` — the version bump and CHANGELOG entry for the release are not in this PR.


itlackey:
Closing — wrong direction. This was `main` -> `release/0.9.8`, inverted from the repo's convention (#883 was `release/0.9.7` -> `main`), because the sprint work was pushed to `main` directly instead of onto a release branch first.

Reopening the conventional way: `release/0.9.8` reset to `main`, carrying the version bump and CHANGELOG entry, PR'd into `main`.

Issue #919, opened 2026-09-03
Editing a workflow breaks its in-flight runs despite the plan being frozen


itlackey:
### Summary

`akm workflow run <ref>` fails permanently on an active run if the workflow source was edited after that run started, reporting `WORKFLOW_IR_VERSION_UNSUPPORTED`. Since the run already froze its own plan at start, editing the source should not be able to invalidate it — the frozen copy is exactly what makes this safe.

### Version

`akm-cli` 0.9.10, backend `cron`, Linux.

### What happened

1. Started a run from an authored workflow, executing part way with `--max-steps`:

   ```sh
   akm workflow run workflows/homelab-health-review --max-steps=4
   ```

   Steps 1–4 completed; the run parked at step 5.

2. Edited the workflow source (`~/akm/workflows/homelab-health-review.md`) — added a field to one step's output schema and extended two step bodies.

3. Ran the ref again, expecting a fresh run:

   ```sh
   akm workflow run workflows/homelab-health-review
   ```

   ```json
   {
     "ok": false,
     "error": "Workflow run 2961f625-... was frozen as workflow plan irVersion 111; pre-irVersion-5 plans cannot execute after the 0.9.2 upgrade. Complete them before upgrading, or run 'akm workflow abandon 2961f625-...' and start a new run from the authored workflow. 'akm workflow status' and 'akm workflow list' still work on this run.",
     "code": "WORKFLOW_IR_VERSION_UNSUPPORTED",
     "hint": "Abandon the run with `akm workflow abandon <id>`, then start it again from the workflow source — pre-0.9.2 frozen plans are not re-executable."
   }
   ```

The run was created minutes earlier on 0.9.10 — no version upgrade happened between creating it and resuming it.

### Two problems

**1. Editing a workflow invalidates its in-flight runs, which defeats the purpose of freezing a plan.**

The docs describe `workflow plan` as freezing the canonical step graph so a run is reproducible. If the plan is frozen at run start, an edit to the source afterwards should be irrelevant to that run — it should keep executing its frozen copy, and only *new* runs should pick up the edit. Instead the edit appears to bump the stored `irVersion` (reported as `111`) and the run becomes permanently unexecutable. This is easy to hit whenever a workflow is being iterated on, which is exactly when runs are most likely to be open.

**2. `akm workflow run <ref>` silently resumes an existing active run instead of starting a new one.**

[...]

Issue #918, opened 2026-09-03
akm config set prints the whole config with no ok/status field, so success and failure look alike to a caller


itlackey:
Small, but it cost me two wrong conclusions in a row while scripting a config change.

## What happens

On failure, `akm config set` returns a clean envelope:

```json
{
  "ok": false,
  "error": "Invalid value for engines.lab:\n  - engines.lab.apiKey: apiKey must be $VAR or ${VAR}",
  "code": "INVALID_FLAG_VALUE",
  "hint": "Run `akm <command> --help` to see accepted values."
}
```

On **success**, it prints the entire resulting config — hundreds of lines, no `ok`, no `shape`, no status field:

```json
{
  "configVersion": "0.9.0",
  "semanticSearchMode": "auto",
  "registries": [ ... ]
```

## Why it matters

A caller checking `.ok` sees `undefined` on success and reports failure. I did exactly that, twice: my wrapper printed `FAILED` for two `config set` calls that had both worked, and I only caught it by dumping the raw output. The asymmetry is the trap — the failure path is well-shaped, so it is natural to assume the success path matches.

It is also a lot of output for a setter. `akm config set` on a large config emits the whole document to say "done".

## Suggestion

Make the success envelope match the failure one:

```json
{ "ok": true, "shape": "config-set-result", "key": "engines.lab" }
```

...and leave the full document to `akm config list`, which is what it is for. If echoing the result is deliberate, adding `"ok": true` alongside it would still be enough for callers to branch on.

Other `akm` commands I used in the same session (`proposal extract`, `health`, `search`) all carry `ok` and `shape` consistently, so this reads as an oversight in one command rather than an intended difference.

itlackey:
**Confirmed, plus a second call site I missed and one correction to my own report.**

## Verified

`src/commands/config-cli.ts:141-144` — on success, `set` calls `output("config", listConfig(updated))`, which routes through the passthrough shape handler (`src/output/shapes/passthrough.ts`). That handler stamps `shape` and `schemaVersion` but **never `ok`**. The failure path is a structurally different envelope built from a thrown `ConfigError`/`UsageError`, caught before `output()` is reached — two disconnected code paths, which is exactly why the shapes disagree.

## I undercounted: `unset` has the identical bug

`config-cli.ts:161` does the same `output("config", listConfig(updated))` on success. Any fix should cover both, or the asymmetry just moves.

## Correction to my own issue text

[...]

Issue #915, opened 2026-09-03
The claude-code -> claude harness rename orphaned extract_sessions_seen with no migration


itlackey:
The harness rename in 0.9.2 left the per-session extraction ledger keyed under the old name, with nothing to carry it forward.

## Observed

`~/.local/share/akm/state.db`, table `extract_sessions_seen` (PK `harness, session_id`):

```
claude-code | no_candidates      4374        opencode | no_candidates       1019
claude-code | skipped              848       opencode | skipped              182
claude-code | candidates_queued    581       opencode | candidates_queued     16
claude      | (1 row, from a manual run today)
```

5,803 rows stranded under `claude-code`. Under `claude`, the ledger is effectively empty.

## Consequences

Both directions are wrong, and they pull opposite ways:

- **Re-extraction.** Every session that was already processed under `claude-code` is now unseen under `claude`, so it is a candidate for reprocessing — LLM cost and duplicate proposals for work already done.
- **History.** The 581 `candidates_queued` and 4,374 `no_candidates` outcomes are no longer attributable to the harness that produced them, so anything reasoning over extraction history (fail rates, coverage, "have I seen this session") silently sees a machine with no past.

In this particular case the orphaning was *convenient* — it is why five days of missed sessions are still re-runnable — but that is luck, not design.

## Suggestion

A one-shot migration renaming `harness = 'claude-code'` to `'claude'` in `extract_sessions_seen` (and anywhere else that column is persisted), gated the way other akm migrations are. If the rename is meant to be a hard boundary, then say so somewhere — but silently splitting a table's key space across a rename is the kind of thing that is invisible until someone queries it.

More generally: the harness id is a persisted key, so renaming one is a schema change, not a string change.

itlackey:
**Confirmed, with the causing commit identified and the migration shape settled.**

## The commit

`35fd98167e9de743d577cc67ec98944b40dbd523` — *"refactor(sessions): delete identity compatibility"* — removed `HARNESS_BY_ANY_ID`, `normalizeHarnessId` and `denormalizeRuntimeIdentity` from `src/integrations/harnesses/index.ts`. The doc comment it deleted said that bridge existed specifically to keep **"existing persisted configs that say 'claude-code' working."**

[...]

Issue #914, opened 2026-09-03
akm health reports pass for an LLM engine that is refusing connections, and pass for extraction that is not running


itlackey:
On a machine where extraction had harvested nothing for ~8 weeks because the configured engine was down, `akm health` reported two passes covering exactly that subsystem.

## Observed

```
default-llm-engine = pass | LLM engine "default" is configured
session-extraction = pass | Session extraction not active (feature disabled or no harness available)
```

Ground truth at that moment:

```
$ curl -m 8 http://192.168.0.99:1234/v1/models
  connection refused                       (host up, service down)

$ akm proposal extract --type claude --since 30d --dry-run
  discovered: 25 | processed: 0 | skipped: 25 | skip reasons: {'llm_unavailable': 25}
```

## The two problems

**1. `default-llm-engine` checks configuration, not reachability.** Its message is honest — "is configured" — but as a *health* check it answers a question nobody is asking. Whether a `llm.endpoint` string exists in a config file is not health; whether the endpoint answers is. A `GET /v1/models` (or a HEAD, or a cheap 1-token completion) with a short timeout would have caught this instantly, and this check is the natural place for it.

**2. `session-extraction`'s "not active" is reported as `pass`.** Reading "Session extraction not active (feature disabled or no harness available)" as healthy means a machine that has stopped harvesting looks the same as one deliberately not configured for it. If akm cannot tell those apart, `unknown` communicates that honestly; `pass` does not.

Both of these are the same shape as itlackey/akm#912: the signals that exist report success while the subsystem does nothing.

## Suggestion

- Give `default-llm-engine` an actual reachability probe (bounded, cached — it runs often), and report `warn`/`fail` with the connection error when it does not answer.
- Make `session-extraction` report `unknown` when it cannot distinguish "off on purpose" from "broken", or better, derive it from the recent `extract_sessions_seen` outcomes — the ledger already knows that 25/25 recent sessions skipped with `llm_unavailable`.

itlackey:
**Both sub-claims verified at source, and point 2 is materially worse than I filed.**

## Point 1 — no reachability check exists anywhere in health

I grepped the entire `src/commands/health/` tree for `fetch(` / `http.request` / any network call: **there are none**. `runConfiguredEngineProbe` (`checks.ts:114-280`) for a plain `kind:"llm"` engine only calls `credentialAvailable()` — an env-var-presence test.

[...]

PR #920, opened 2026-09-03
release: 0.9.12

itlackey:
## Summary

Release 0.9.12: eight issues from the milestone, plus a repo-wide audit of defensive refusals.

### Milestone issues

- #912 — `akm proposal extract` aggregates infrastructure skip reasons into `warnings[]` (`25 of 25 sessions skipped: llm_unavailable (engine "default")`) and adds a `skipReasons` count map. `ok` keeps meaning "the command ran".
- #913 — the extract envelope carries `engine` and `engineKind`, every `sessions[]` entry and the `extract_sessions_seen` ledger metadata carry `engine`. `akm health`'s `active-improve-strategy` names the engine each improve process resolved to.
- #914 — `default-llm-engine` / `configured-engines` send one bounded (3 s) `GET` to the endpoint's `/models` route, once per distinct endpoint per invocation, with `--no-probe` for offline hosts. `session-extraction` derives its verdict from the last seven days of the `extract_sessions_seen` ledger only.
- #915 — state migration `027-extract-sessions-seen-harness-rename` moves `extract_sessions_seen` and `workflow_runs.agent_harness` rows off `claude-code` (conflict-tolerant); a test pins the persisted harness-id list so the next rename cannot ship without a migration.
- #918 — the passthrough stamp adds `ok: true` to every success envelope that has no `ok` of its own; commands whose exit code grades the outcome (`task sync`, `task sync --dry-run`, `task prune`, `workflow run`, `upgrade`, `migrate *`) emit through one `outputWithExitCode` helper that sets `ok` and the exit code from the same value. `--silent` still prints nothing.
- #919 — `workflow run <ref>` reports `resumed: true` and a "resuming existing run … pass --new" line when it attaches to an active run; `--new` starts a fresh run; the unsupported-plan message distinguishes a plan frozen by a newer akm from a pre-irVersion-5 one; `status`, `abandon`, `resume`, and `run <id>` accept unique run-id prefixes of eight or more characters. The reported `irVersion 111` did not reproduce; a regression test pins that editing a source leaves its in-flight run executable.
- #925 — `akm agent` no longer rejects a prompt containing `{{` or `}}`. The portable-template construct scan is gone, and inline agent content dispatches verbatim.
- #928 — a state.db whose migration ledger is newer than this binary's now warns once and degrades instead of refusing to open. Note the companion `runFtsQuery … no such column: e.entry_json` noise comes from an older *bundled* akm reading a newer index; that resolves when the bundled binary moves to 0.9.12+, not from this change.

### Guard audit

Every defensive refusal in the codebase was judged against three tests: has it demonstrably helped a real user, does its failure mode cost less than the hazard it guards, and is the hazard already gated behind a deliberate human command. Refusals that failed were removed, or degraded to a warning, in that order of preference. Machinery preventing data loss or corruption — atomic writes, backups, write-path validation, path containment — was out of scope and is unchanged.

The user-visible effect is that akm stops aborting on conditions it can survive: config load no longer bricks every command over one bad key; a newer or unfamiliar migration ledger degrades; free text is no longer validated as code; version skew is no longer treated as corruption; deliberate commands (`--since`, `--force`) are no longer second-guessed; symlinks are followed on read paths with realpath containment doing the safety work; and blocks that had no escape hatch got one (`env run --allow-insecure`) or were removed.

Guards that passed the three tests were kept and documented, including the registry-URL credential inspection limit (removing it let a hidden url-encoded credential reach a live request), the third-party dangerous-key block, `MAX_SCHTASKS_TRIGGERS`, the cron line-length limit (a truncated cron line executes a partial command — the over-length case now spills into a wrapper script instead of refusing), and every path-containment check.

Three defects surfaced during the merge that no single area branch could have caught, each root-caused rather than worked around: a collision test asserting a throw that two independently-correct changes removed (verified by capturing the dispatcher payload that the merged behaviour serves the `.md` winner, never the stale `.yml`); a health probe whose exception-based detection became dead code, silently reporting a clean pass for a real misconfiguration; and the same ahead-blocks-noop bug in a sibling file the audit never named.

## Test plan

- `bun run lint`, `bunx tsc --noEmit`, `bun run test:unit`, full `bun run test:integration` green on the candidate.
- Release candidate evidence (exact SHA, Gated CI run, local `./tests/release-check.sh --skip-docker`) is posted as a comment and refreshed on every candidate.

Publishing is a separate step (Release workflow with `version: 0.9.12`) and is not part of this PR.

The remaining milestone items (#921, #922, #923, #924, #926, #927, #929, #930) are moved to 0.9.13; they are not addressed here.

## Checklist

- [x] Tests pass (`bun test`)

https://claude.ai/code/session_01GdS26JcJ5A9Hqxxe3jo31x


itlackey:
## Release-candidate evidence for 0.9.12

- **Candidate SHA:** `169fd6a2c68dbb5984d8e253347f88256e3da0a4` (`release: 0.9.12`)
- **Exact-SHA Gated CI:** https://github.com/itlackey/akm/actions/runs/33736506103 — `workflow_dispatch` with `candidate_sha` set to that SHA; Resolve Candidate, Semantic Search, Docker Install, Native Scheduler Linux/macOS/Windows, and Release Candidate Evidence all succeeded, run attempt 1.
- **Local `./tests/release-check.sh --skip-docker`** on the same commit: `Release validation passed` — unit 6467 pass / 5 skip / 0 fail (474 files), integration 3490 pass / 55 skip / 0 fail (288 files), package acceptance, install/setup regression, and the Linux standalone artifact all green.
- **PR CI on this head:** `check`, `node-smoke (22)`, `node-smoke (24)`, `determinism` all green.

Any further commit on this branch creates a new candidate and invalidates this evidence. Publishing (Release workflow, `version: 0.9.12`) is a separate step and has not been run.

---

itlackey:
## Release-candidate evidence for 0.9.12 (supersedes the comment above)

Two commits landed after the first candidate: `2d3d4c6f` restores `engineKind` on the extract envelope (`llm` | `sdk` | `agent`), and `acb04b76` makes `session-extraction` read the ledger only and collapses the two run-id resolvers into one.

- **Candidate SHA:** `acb04b764d48d6562962bf79ee58d6e861aaedac`
- **Exact-SHA Gated CI:** https://github.com/itlackey/akm/actions/runs/33767602273 — `workflow_dispatch` with `candidate_sha` set to that SHA; Resolve Candidate, Semantic Search, Docker Install, Native Scheduler Linux/macOS/Windows, and Release Candidate Evidence all succeeded, run attempt 1.
- **Local `./tests/release-check.sh --skip-docker`** on the same commit: `Release validation passed` (full unit and integration targets, package acceptance, install/setup regression, Linux standalone artifact).
- **PR CI on this head:** `check`, `node-smoke (22)`, `node-smoke (24)`, `determinism` all green.

The evidence in the earlier comment (SHA `169fd6a2`) is void. Publishing (Release workflow, `version: 0.9.12`) has not been run.

---

itlackey:
## Release-candidate evidence for 0.9.12 (supersedes the earlier comment)

- **Candidate SHA:** `0729b30c82070d88d1ea6a83880617322236b7ee` (head of `release/0.9.12` after the cleanup pass)
- **Exact-SHA Gated CI:** https://github.com/itlackey/akm/actions/runs/33772038186 — `workflow_dispatch` with `candidate_sha` set to that SHA; Resolve Candidate, Semantic Search, Docker Install, Native Scheduler Linux/macOS/Windows, and Release Candidate Evidence all succeeded, run attempt 1.
- **Local `./tests/release-check.sh --skip-docker`** on the same commit: `Release validation passed` — unit 6407 pass / 5 skip / 0 fail (474 files), integration 3489 pass / 55 skip / 0 fail (288 files), package acceptance, install/setup regression, and the Linux standalone artifact all green.
- **PR CI on this head:** `check`, `node-smoke (22)`, `node-smoke (24)`, `determinism` all green.

Any further commit on this branch creates a new candidate and invalidates this evidence. Publishing (Release workflow, `version: 0.9.12`) is a separate step and has not been run.

---

itlackey:
## Release candidate evidence — `c23613bf00f595baf62e7fd395b9ecd9f2031658`

This supersedes the earlier evidence comments; they named heads that no longer exist on this branch.

### Gated CI — [run 33803147751](https://github.com/itlackey/akm/actions/runs/33803147751), exact SHA, all 7 jobs green

| Job | Result |
| --- | --- |
| Gated / Resolve Candidate | success |
| Gated / Release Candidate Evidence | success |
| Gated / Docker Install | success |
| Gated / Semantic Search | success |
| Gated / Native Scheduler / Linux | success |
| Gated / Native Scheduler / macOS | success |
| Gated / Native Scheduler / Windows | success |

An earlier dispatch (run 33803041546) failed at Resolve Candidate because the `candidate_sha` input was given as the short `c23613bf`, which the checkout resolved as a branch ref (`refs/heads/c23613bf*`) and could not find. The candidate was never checked out, so nothing was tested; re-dispatched with the full 40-character SHA. Not a code failure and not a flake.

### Local gates on the same SHA

| Gate | Result |
| --- | --- |
| `bunx tsc --noEmit` | clean |
| `bun run lint` | clean (exit 0) |
| `bun run test:unit` | 6506 pass / 5 skip / 0 fail (476/476 files) |
| `bun run test:integration` | 3516 pass / 55 skip / 0 fail (291/291 files) |
| `./tests/release-check.sh --skip-docker` | Release validation passed |

Standard PR checks (`check`, `determinism`, `node-smoke 22`, `node-smoke 24`) are also green on this head.

### What landed since the last evidence comment

A repo-wide audit of defensive refusals, merged from six area branches (core, storage, integrations, improve, workflows, cli). Each refusal was judged against three tests: has it demonstrably helped a real user, does its failure mode cost less than the hazard, and is the hazard already gated behind a deliberate human command. Machinery preventing data loss or corruption — atomic writes, backups, write-path validation, path containment — was out of scope and is unchanged.

Guards that passed the three tests were kept and documented, including the registry-URL credential inspection limit (removing it let a hidden url-encoded credential reach a live request), the third-party dangerous-key block, `MAX_SCHTASKS_TRIGGERS`, the cron line-length limit (a truncated cron line executes a partial command — the over-length case now spills into a wrapper script instead), and every path-containment check.

Three defects found during the merge that no single branch could have caught:

- `tests/integration/workflows/workflow-source-collision.test.ts` asserted a throw that two independently-correct changes removed. Verified by capturing the dispatcher payload that the merged behaviour serves the `.md` winner and never the stale `.yml` sibling — better than the refusal it replaced. Test rewritten to pin both halves.
- Finding 14 made a health probe's exception-based detection dead code, so `selected-model-aliases` silently reported a clean pass for a real misconfiguration. Restored via an explicit `unmappedForEngine` discriminator.
- `src/sources/providers/git-install.ts` carried the same ahead-blocks-noop bug as its sibling, in a file the audit never named.

`registry-pinned-transport.test.ts`'s response-header timeout was raised from 25 ms to 250 ms: the server never responds so the timeout always fires, but under sharded load the helper spawn alone could exceed 25 ms and fail the assertion before the helper existed.

Publishing remains a separate step (Release workflow with `version: 0.9.12`) and is not part of this PR.

---

Issue #930, opened 2026-09-03
Evaluate FTS5 field weights after stable score normalization; ship only a measured non-regression


itlackey:
## What

`src/storage/repositories/index-fts-repository.ts:117`:

```sql
bm25(entries_fts, 0, 10.0, 5.0, 3.0, 2.0, 1.0) AS bm25Score
```

against the FTS5 schema at `src/storage/repositories/index-schema.ts:253-261`:

| column | `name` | `description` | `tags` | `hints` | `content` |
|---|---|---|---|---|---|
| weight | **10x** | **5x** | 3x | 2x | **1x** |

Document body prose is weighted an order of magnitude below the title and 5x below the description, then further diluted by BM25's length normalization over a long document.

Still present on `main` at 0.9.11.

## Why this is the wrong shape for what akm does

akm's core claim is finding the *fact you need* inside your knowledge, not finding a document by its name. But a fact stated in a document's opening sentence receives a large ranking boost, while the identical fact stated three paragraphs in competes at 1x. Users keep their content in bodies.

## Impact, measured

From the LongMemEval n=200 round on 0.9.10, where each session is one indexed document and `description` is the synthesized first ~250 characters:

- **4 of the 8** `knowledge-update` questions that akm failed and full-context answered correctly were traced to this by reading the source text: in every case the *retrieved* session states its topic in sentence one, and the *missed* session opens on an unrelated topic and mentions the needed fact only as an aside several turns in.
- `knowledge-update` scored 0.727 vs 0.909 for a full-context baseline, and unlike other categories that deficit is **not** confounded by question difficulty — the baseline scores ~0.89-0.92 regardless of which coverage bucket akm's retrieval puts a question in, so the coverage effect is attributable to retrieval.
- Partial-coverage `knowledge-update` questions averaged 1.11 results retrieved, i.e. the needed document was never a candidate at all rather than being ranked too low.

## Suggested fix

Raise the `content` weight substantially relative to `name`/`description`, and/or stop deriving ranking-relevant `description` weight from what is effectively "the first N characters of the body". A document's opening should not be a proxy for its relevance.

This needs measuring rather than guessing — the right weights are an empirical question, and the change affects every akm query.

## How to verify without spending judge budget

```sh
bin/probe <version>   # in akm-eval; deterministic, LLM-free, minutes
```

[...]

PR #931, opened 2026-09-03
fix(search): top up the FTS tiers instead of returning from the first hit (#929)

### LINKED ISSUE #929: searchFts short-circuits the FTS cascade, starving multi-fact queries of candidates
## What

`searchFts` (`src/storage/repositories/index-fts-repository.ts:58-82`) tries exact-AND → prefix-AND → relaxed-OR and **returns on the first tier that yields any hit at all**, never topping up the remaining `limit` slots from the looser tiers:

```ts
const exactResults = runFtsQuery(db, plan.exact, "exact", limit, ...);
if (exactResults.length > 0) return exactResults;      // returns on ONE hit
if (plan.exactPrefix) { ... if (prefixResults.length > 0) return prefixResults; }
return plan.relaxed ? runFtsQuery(db, plan.relaxed, "relaxed", limit, ...) : [];
```

If a query's strict conjunctive form matches exactly one document, that single document is all the caller ever receives — even when a second, genuinely relevant document phrased slightly differently would have surfaced from the prefix or relaxed tier. The tiers are alternatives, not a progressive fill.

Still present on `main` at 0.9.11.

## Impact, measured

From the LongMemEval n=200 round on akm-cli 0.9.10 (`akm-eval` `runs/RESULTS-n200-0.9.10.md`):

- **23 of 200 questions** returned an incomplete result set that was **not** topK-bound — the search stopped early, it did not run out of slots. 22 of those 23 need more than one evidence document.
- The `knowledge-update` questions with partial coverage averaged **1.11 results retrieved in total**. That is candidate starvation, not a ranking cutoff.
- The result-count distribution is bimodal and gives the mechanism away: `{1: 35, 2: 12, 3: 1, 4: 0, 5: 152}`. **Zero queries returned exactly 4.** 76% fall through to OR and saturate `topK`; the rest get a small exact-match set. A relevance threshold would produce a smooth curve — this is a switch.

Three independent investigations of that run converged on this line from different starting points.

Any akm query whose strict form happens to match one document silently stops there, for every user. The failure is invisible: results come back, they are relevant, and there is nothing to indicate that better candidates were never considered.

## Suggested fix

Union the tiers up to `limit` instead of returning early, deduping by path the way the ranker already does. Same FTS table, same BM25 weights, same downstream ranker — only the candidate pool stops being truncated.

## How to verify without spending judge budget

`akm-eval`'s deterministic probes cover exactly this and cost nothing:

```sh
bin/probe <version>   # graded against committed reference values, LLM-free
```

Watch `recallAtK` and `avgResultsReturned`. Expected: `avgResultsReturned` rises toward `topK`, recall rises, and per the same run's analysis precision may fall without hurting answers (noise correlates with correctness at r = -0.078 there; recall correlates at +0.42).

Related: #930

itlackey:
Closes #929 — **draft: do not merge yet.** Two tests fail and both are correct to; see the blocker below.

Rebased onto `release/0.9.12` (clean, no conflicts) and retargeted from `main`.

## The fix

`searchFts` returned from whichever tier first produced a hit, so a query whose strict conjunctive form matched one document got exactly that document and the looser tiers never ran. Nothing surfaced the truncation.

The tiers now fill a shared pool. Two things the naive union got wrong — both caught by **existing** tests, not by inspection:

**Precision.** An unfiltered OR top-up appends documents matching a single term of a deliberately precise query: `"deploy kube"` surfaced a *deploy docker* document on the shared word "deploy" (`tests/integration/fuzzy-search.test.ts`). The relaxed tier now plays two distinct roles — as a **fallback** (nothing matched conjunctively) it stays unfiltered, because any hit beats none; as a **top-up** it requires a document to carry at least two query terms. A two-term query therefore collapses to the conjunctive result, which is right for precise input, while a sentence-shaped question still admits related documents.

**A downstream invariant.** `normalizeFtsScores` reads `results[0]` as best and `results[last]` as worst to build its min-max range — it requires a bm25-sorted array. Concatenating tiers silently corrupted every normalized score. The merged pool is re-sorted before truncation, each result keeping its own `lexicalMatch` so a promoted relaxed row still meets the ranker's relaxed-tier ceiling. A full exact tier short-circuits before any of this, so the common case is byte-identical to today.

6 new tests in `tests/integration/db/db-scoring.test.ts` cover top-up, the two-token floor, ordering, tier labels, fallback-stays-unfiltered, dedupe, limit, and empty.

## Blocker — why this is a draft

**Enlarging the candidate pool inflates normalized fts scores.** `normalizeFtsScores` derives its floor from the *worst candidate in the set*, so adding candidates raises everything above it and pushes more results into the score clamp.

Measured on `release/0.9.12`, query `"database outage recovery"`, `incident-checklist`:

| | score | rank | full score array |
|---|---|---|---|
| `maxHops: 1` | `1` | 3 | `[1, 1, 1, 0.549]` |
| `maxHops: 2` | `1` | 3 | `[1, 1, 1, 0.549]` |

**Three of four results are pinned at exactly `1.0`.** The graph boost is still applied — it is invisible because the score saturates. `graph-boost-ranking` asserts `twoHopScore > oneHopScore` and can no longer observe it. **This is the real issue: retrieval improved and ranking got less expressive.**

The second failure, `improve-memory`'s belief-state test, finds a newly-visible legitimate parent memory (`memories/deploy`) between its two expected hits. Its actual invariant — the current derived memory outranks the contradicted one — still holds; `[1]` encoded an incidental fact. Stale assertion.

I deliberately did not weaken either test. The first is correctly reporting a genuine regression in ranking behaviour.

**This should not merge until fts scoring stops deriving its floor from the worst candidate in the set** — absolute scoring, per-tier normalization, or a fixed reference scale. That is a separate change with its own blast radius.

## Verification, on `release/0.9.12`

Full `tests/integration/` suite, this branch vs. the base with the commit removed:

| | pass | fail |
|---|---|---|
| `release/0.9.12` baseline (scratch worktree, no change) | 3517 | 3 |
| this branch | 3523 | 5 |

`+6 pass` is exactly the 6 new tests. **Exactly 2 failures are attributable to this change** — the same two documented above, unchanged from when the work was based on `main`.

The other **3 failures are pre-existing on `release/0.9.12`** and unrelated to search, confirmed by running the baseline worktree: two `workflow source canonical-ref collisions` cases and one `resolveJudge` gated-workflow freeze. Flagging them because `release/0.9.12` is not currently green, not because this PR touches them.

- `bunx tsc --noEmit`: **0 errors** on both the branch and the baseline.
- Targeted: the four relevant files are 40 pass / 0 fail at baseline, 46 pass / 2 fail on the branch.
- Free deterministic check available in `akm-eval`: `bin/probe <version>` grades recall/precision on two frozen corpora with no LLM.

https://claude.ai/code/session_01SMms9XDaLaqS1M3bcpJ6df

### DIFF
--- a/src/storage/repositories/index-fts-repository.ts
+++ b/src/storage/repositories/index-fts-repository.ts
@@ -65,20 +65,90 @@ export function searchFts(
   const plan = buildLexicalQueryPlan(query);
   if (!plan.exact) return [];
 
-  // Try the exact AND query first
+  // The tiers TOP UP a shared pool; they are not alternatives (#929).
+  //
+  // This used to return from whichever tier first produced a hit. A query
+  // whose strict conjunctive form matched exactly one document therefore
+  // yielded exactly that one document, and the looser tiers — which would
+  // very often have supplied the rest — never ran. The caller could not tell:
+  // results came back, they were relevant, and nothing indicated that better
+  // candidates were never considered. Measured on LongMemEval, 23 of 200
+  // queries returned a short result set that was NOT limit-bound, 22 of them
+  // on questions needing more than one document.
   const exactResults = runFtsQuery(db, plan.exact, "exact", limit, entryType, excludeTypes);
-  if (exactResults.length > 0) return exactResults;
+
+  // Fast path, byte-identical to the old behaviour: a full exact tier needs no
+  // top-up, so the looser queries never run and nothing is re-ordered.
+  if (exactResults.length >= limit) return exactResults;
+
+  const pool: DbSearchResult[] = [];
+  const seenEntryIds = new Set<number>();
+  const addTier = (tierResults: DbSearchResult[]): void => {
+    for (const result of tierResults) {
+      if (seenEntryIds.has(result.id)) continue;
+      seenEntryIds.add(result.id);
+      pool.push(result);
+    }
+  };
+
+  addTier(exactResults);
 
   if (plan.exactPrefix) {
-    const prefixResults = runFtsQuery(db, plan.exactPrefix, "prefix", limit, entryType, excludeTypes);
-    if (prefixResults.length > 0) return prefixResults;
+    addTier(runFtsQuery(db, plan.exactPrefix, "prefix", limit, entryType, excludeTypes));
   }
 
-  // One measured relaxation only after both conjunctive forms miss. This is
-  // still the same FTS table, BM25 weights, candidate collection, and
-  // downstream ranker — merely an OR candidate query for sentence-shaped
-  // input whose filler terms prevented a strict hit.
-  return plan.relaxed ? runFtsQuery(db, plan.relaxed, "relaxed", limit, entryType, excludeTypes) : [];
+  // The relaxed OR pass plays two different roles, and conflating them is a
+  // precision bug.
+  //
+  //  - As a FALLBACK (nothing matched conjunctively), any single-term hit
+  //    beats returning nothing. Unfiltered, exactly as before.
+  //  - As a TOP-UP (a thin conjunctive match left room), an unfiltered OR
+  //    appends documents matching just one term of a deliberately precise
+  //    query. Searching "deploy kube" would surface a doc about "deploy
+  //    docker" purely on the shared word "deploy" — noise the caller did not
+  //    ask for, promoted into a result set that was already correct.
+  //
+  // So a top-up requires a document to carry at least two of the query's
+  // terms. For a two-term query that collapses to the conjunctive result and
+  // adds nothing, which is right for precise input; for a sentence-shaped
+  // question it still admits genuinely related documents while rejecting
+  // single-common-word coincidences.
+  if (plan.relaxed) {
+    const relaxedResults = runFtsQuery(db, plan.relaxed, "relaxed", limit, entryType, excludeTypes);
+    const isTopUp = pool.length > 0;
+    addTier(isTopUp ? relaxedResults.filter((r) => countMatchedTokens(r, plan.tokens) >= 2) : relaxedResults);
+  }
+
+  // Re-sort the merged pool by bm25 before truncating. This is REQUIRED, not
+  // cosmetic: `normalizeFtsScores` (indexer/search/ranking.ts) reads
+  // `results[0]` as the best score and `results[last]` as the worst to build
+  // its min-max range, so handing it a tier-concatenated array silently
+  // corrupts every normalized score. bm25 is the same function over the same
+  // table and weights in all three tiers, and rewards matching more of the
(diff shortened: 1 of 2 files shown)


itlackey:
## Do not merge. Measured on the frozen probe corpora, this fix does not work.

I ran `akm-eval`'s deterministic probes against this branch and against stock `release/0.9.12` as a control. The control matches the committed reference values exactly on all 8 metrics, so every delta below is attributable to this commit alone.

| metric | stock 0.9.12 | with this fix | change |
|---|---|---|---|
| locomo `evidenceRecall@5` | 0.590 | 0.590 | **0** |
| locomo `recallAtK` | 0.497917 | 0.497917 | **0** |
| locomo `precisionAtK` | 0.232917 | **0.147500** | **−37%** |
| longmemeval `evidenceRecall@5` | 0.950 | 0.950 | **0** |
| longmemeval `recallAtK` | 0.950000 | 0.950000 | **0** |
| longmemeval `precisionAtK` | 0.676667 | **0.275000** | **−59%** |

**Recall does not move by a single digit on either corpus.** Precision collapses. The change returns strictly more documents and not one of them is evidence — it is pure noise addition.

## Why the premise was wrong

The issue assumed missing evidence was a *candidate-count* problem: the cascade stopped early, so the needed document never entered the pool. The measurement says otherwise. Adding candidates from the looser tiers surfaces other documents, never the missing ones.

That is consistent with #930: the evidence these queries miss is buried in document **bodies**, and `content` is weighted `1x` against `name` at `10x` and `description` at `5x`. Such a document does not rank into the top 5 no matter how large the candidate pool gets. Generating more candidates cannot fix a ranking problem.

So #929 looks like a red herring and **#930 is the real fix.** I had this ranked the wrong way round.

## Disposition

Keeping this as a draft rather than closing it, because the two defects it *did* uncover are real and worth preserving:

- `normalizeFtsScores` requires a bm25-sorted array (it reads `results[0]`/`results[last]` for its min-max range) — undocumented, and silently corrupts every score if violated.
- Enlarging the candidate pool inflates normalized scores into the clamp, so downstream boosts lose expressive range. Demonstrated: `[1, 1, 1, 0.549]` identically under `maxHops: 1` and `maxHops: 2`.

Both need addressing before any change that widens candidate pools, including a topK increase.

Reproduce in `akm-eval`, free and LLM-free, about two minutes:

```sh
bin/probe --cmd '["bun","/path/to/akm/src/cli.ts"]'
```

itlackey:
Closing this draft as a preserved negative result. The implementation remains available on the branch/commit, but the frozen probes showed zero recall gain, 37%/59% precision loss, and score-clamp saturation. #929 is closed as not planned; #933 and #937 now track the validated defects. Reopen only if new evidence overturns that result.

Issue #927, opened 2026-09-03
akm lint reports templated output paths (portfolio-review-<timestamp>.md) as stale-path


itlackey:
### Summary

`akm lint` reports a filename **pattern** as a missing path. A workflow that documents where it writes its report:

```markdown
Write the report into `/home/founder3/akm/reports/`, named
`portfolio-review-` plus a `YYYYMMDDHHMMSS` stamp and `.md`.
```

was fine, but the more natural phrasing:

```markdown
Write to `/home/founder3/akm/reports/portfolio-review-<timestamp>.md`.
```

produces:

```json
{
  "file": "workflows/portfolio-review.md",
  "issue": "stale-path",
  "detail": "nonexistent path: /home/founder3/akm/reports/portfolio-review-<timestamp> (portable form: $AKM_BUNDLE_DIR/reports/portfolio-review-<timestamp>)"
}
```

The path does not exist because it is a template for a filename generated at run time. It never will exist under that literal name.

### Why it is worth fixing

`stale-path` is a genuinely useful check — it caught real broken references elsewhere in my bundle. Its value depends on a clean signal, and workflow documents are exactly where run-time output paths get described, so this false positive lands disproportionately on the files where the check would otherwise be most useful.

The workaround (splitting the directory from the filename in prose) makes the document worse to read in order to satisfy a linter, which is the wrong trade.

### Suggested fix

Skip `stale-path` for a path containing an obvious placeholder. Any of:

- angle-bracket segments — `<timestamp>`, `<name>`, `<id>`
- brace segments — `{stamp}`, `${VAR}`
- a run of date-format characters — `YYYYMMDD`, `HHMMSS`
- a `*` or `?` glob character

Reporting it at a lower severity, or as `templated-path` rather than `stale-path`, would also work — the distinction that matters is "this reference is broken" versus "this reference is parameterised".

### Version

`akm-cli` 0.9.10. `akm lint` on a bundle with 34 flagged findings total.

Issue #926, opened 2026-09-03
Workflow level-2 headings must be step ids, which is only discoverable by breaking compilation


itlackey:
### Summary

Every level-2 heading in a workflow markdown document must exactly match a declared step id. Adding a section of explanatory prose breaks compilation:

```
$ akm workflow plan workflows/portfolio-review
{
  "ok": false,
  "error": "Workflow source has 1 error(s):\n  /home/founder3/akm/workflows/portfolio-review.md:147 — Unexpected level-2 heading \"### How context moves between steps\" on line 147 — no step \"How context moves between steps\" is declared in frontmatter \"steps:\". Level-2 headings must exactly match a declared step id.",
  "code": "WORKFLOW_SOURCE_INVALID"
}
```

The error itself is clear and points at the right line. The problem is that nothing tells you the constraint exists until you trip it, and `##` is the natural heading level for a document section.

### Why it bites

The step bodies in a real workflow are long, and the useful things to write down are often *cross-cutting* rather than belonging to one step — how artifacts move between steps, a shared output contract, environment prerequisites, traps that apply to several steps. `##` is where a writer reaches for those. Demoting them to `###` works but reads oddly, since they are peers of the step sections, not children of the preceding one.

I hit this twice in one session, the second time having already learned it once, because the constraint is invisible while editing.

### Suggested fixes, cheapest first

1. **Say it in the template.** `akm workflow create --print` emits a starter document; a one-line comment there ("level-2 headings are step ids; use `###` for anything else") would prevent most of these.
2. **Allow non-step level-2 sections by convention** — anything before the first step heading, or a small set of reserved names (`## Notes`, `## Overview`, `## Prerequisites`) that the compiler ignores.
3. **Extend the error with the fix**, not just the diagnosis: "…must exactly match a declared step id. To document something that is not a step, use a level-3 heading."

Item 3 costs one sentence and is worth it regardless of the others.

### Version

`akm-cli` 0.9.10.

Issue #924, opened 2026-09-03
Run-lease contention is reported as SQLite corruption (database is locked / disk image is malformed / disk I/O error)


itlackey:
### Summary

When a workflow run lease is already held, `akm workflow run` frequently reports a raw SQLite error instead of the lease condition. Over one session the same underlying situation produced three different messages:

```json
{"ok": false, "error": "database is locked"}
{"ok": false, "error": "database disk image is malformed"}
{"ok": false, "error": "disk I/O error"}
```

None of these is true. `PRAGMA integrity_check` passes on all three databases, and the disk is fine:

```
state.db   ok        (1.5 GB)
index.db   ok        (586 MB)
logs.db    ok        (539 MB)
/dev/nvme1n1p2  1.8T  896G  780G  54% /home     inodes 9% used
```

`database disk image is malformed` in particular reads as data loss. It sent me to integrity checks, backup inspection, and a WAL/SHM review before I found the real cause.

### The real error, only visible with --verbose

```
$ akm workflow run workflows/homelab-spot-check --max-steps=1 --verbose
{
  "ok": false,
  "error": "Workflow run d36025fe-... is already being driven by engine 12424880-... (run lease expires 2026-09-03T06:16:33.002Z). A second `akm workflow run` would race it — wait for that invocation to finish or for the lease to expire.",
  "code": "INVALID_FLAG_VALUE"
}
```

That message is excellent. It should be what a caller sees at default verbosity.

### Two smaller problems in the same area

**Wrong error code.** The lease message is tagged `code: "INVALID_FLAG_VALUE"`. Nothing about a held lease is a flag-validation problem, and a caller branching on `code` cannot distinguish "retry shortly" from "you passed bad input". A dedicated code (`RUN_LEASE_HELD` or similar) would make it programmatically handleable — this is the one error a wrapper script most wants to retry on.

**Orphaned leases.** Twice, an engine process died without releasing its lease or recording progress. The run stayed `status: active` with `currentStepId` set and a live `engineLease`, while `ps` showed no engine:

```
status: active | step: collect-hosts | lease: True | updated: 2026-09-03T06:44:42Z
$ ps -o cmd= -C node | grep 'workflow run'    # nothing
```

The lease does expire on its own, so this self-heals, but during the window `status` asserts the run is being driven when it is not.

### Reproduction

[...]

PR #932, opened 2026-09-04
release: 0.9.13

itlackey:
## Summary

Seven of the nine issues on milestone 0.9.13. Two are deliberately held — see below.

- **#922** — every list-returning command exposes its collection under a stable `results` key alongside its semantic one (`hits`, `items`, `proposals`, …), in every shape including `--shape agent`. Same array, not a copy. Eleven commands covered, not the five named in the issue; `health` and `task doctor` are excluded because they carry several heterogeneous collections and picking one would be arbitrary.
- **#917** — `apiKey` accepts `secret://<name>` alongside `$VAR`/`${VAR}`, resolved through the existing store-backed resolver, on both the engine and embedding paths. Literal keys in `config.json` are still refused; an unresolvable reference fails loudly, naming the reference and never the secret.
- **#921** — `proposal drain` reports a `failed[]` naming each refused proposal and why, instead of `failed: 0` alongside five stderr failures. `--dry-run` applies the same stale-target check as the real run, so the two stop disagreeing. `akm improve`'s triage pre-pass reports the same count. The refusals themselves were always correct and are unchanged.
- **#923** — a step whose returned object does not match its declared `outputSchema` now says so, naming the step and the specific problem. A warning: the run continues and its status is unchanged, because a workflow is a guide rather than a contract.
- **#924** — a held run lease reports as itself at default verbosity with a dedicated `RUN_LEASE_HELD` code, rather than as `database is locked` / `disk image is malformed` / `disk I/O error`. Genuine SQLite errors still report as themselves.
- **#927** — `akm lint` recognises templated output paths (`<timestamp>`, `{stamp}`, `${VAR}`, `YYYYMMDD`, globs) as parameterised rather than broken. Genuinely missing literal paths are still reported.
- **#926** — the `workflow create` template states the level-2-heading rule, and the compiler's rejection now carries the remedy rather than only the diagnosis.

## Held back: #929 and #930

Both change retrieval and are measurable only against the external LongMemEval corpus, which cannot be run in this environment, so they are being evaluated together rather than shipped blind. #930 was deferred by decision before work started; #929 was implemented and then held on evidence. The work is preserved on `issue/0.9.13-search` at `8c61f1f0`.

Investigation after this PR opened found the two issues are **the same failure — candidate starvation, not ranking**. Both cite the same "1.11 results retrieved" figure, and #930's own text says the needed document "was never a candidate at all rather than being ranked too low". Because `runFtsQuery` ends in `ORDER BY bm25Score, e.id ASC LIMIT ?`, #930's 10:1 title-to-body weighting gates which rows are returned at all, not merely their order — it is a recall lever, not a ranking preference.

It also identified a defect underneath both, which neither issue names. `normalizeFtsScores` (`ranking.ts:71-86`) reads `results[0]` and `results[length-1]` as best/worst, assuming a sorted array — that assumption is what forced synthetic bm25 offsets in the #929 attempt — and then min-max normalizes over the candidate set, so a document's score depends on what else matched. Measured: three additional weak candidates compress leader separation by 85% and push mid-pack scores toward the 1.0 clamp in `db-search.ts:577`. `ranking-contributors.ts:165-186` already documents this pathology and works around it with hard-coded belief-state score ceilings.

The two failing tests are not equivalent, and an earlier revision of this description overstated them as both being ranking-quality tradeoffs:

- `graph-boost-ranking` "maxHops=2" is a real defect exposed by the change: it fails with `oneHopScore` and `twoHopScore` **both exactly `1`** — the base score saturates the clamp, leaving the boost no headroom to differentiate. Reachable today with a narrower pool, just less often.
- `improve-memory` "prefers the current derived memory" **still passes its headline assertion at position 0**. Only the runner-up changed, because a previously starved parent memory now legitimately surfaces. That assertion encodes the starvation and should be rewritten, not treated as a regression.

Fix order, once the corpus is available: repair the normalization first (it is the prerequisite, and it lets the synthetic offsets be deleted), then #929's union becomes safe, then #930's weights measured as a recall change.

## Test plan

- `bun run lint`, `bunx tsc --noEmit`, `bun run test:unit`, full `bun run test:integration` green on the candidate.
- Release candidate evidence (exact SHA, Gated CI run, local `./tests/release-check.sh --skip-docker`) posted as a comment and refreshed on every candidate.

Do not merge until the evidence comment names the current head. Publishing is a separate step (Release workflow with `version: 0.9.13`) and is not part of this PR.

## Checklist

- [x] Tests pass (`bun test`)

https://claude.ai/code/session_01GdS26JcJ5A9Hqxxe3jo31x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,67 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [0.9.13] - 2026-09-04
+
+### Added
+
+- **A stable `results` alias on every list-returning command (#922).** `search`,
+  `curate`, `proposal list`, `bundle list`, `env list`, `secret list`,
+  `registry search`, `registry list`, `workflow list`, `task history` and
+  `log list` each keep their existing semantic key (`hits`, `items`,
+  `proposals`, …) and now also expose the same array as `results`, in every
+  shape including `--shape agent`. It is the same array, not a copy. A caller
+  reading `hits` from a `curate` response previously got nothing back and could
+  reasonably read that as "no results" — there was no error and the `summary`
+  alongside it still reported that results were selected. Commands carrying
+  several heterogeneous collections (`health`, `task doctor`) are deliberately
+  excluded.
+- **Engine and embedding credentials can resolve from the secret store
+  (#917).** `apiKey` accepts `secret://<name>` alongside `$VAR` / `${VAR}`,
+  resolved through the existing store-backed resolver. A detached SessionEnd
+  hook or a cron job — the two contexts least able to supply an environment
+  variable and most likely to need extraction — no longer requires editing the
+  login environment for a value akm already stores. Literal keys in
+  `config.json` are still refused, and an unresolvable reference fails loudly,
+  naming the reference and never the secret.
+- **`akm proposal drain` reports what failed (#921).** The envelope carries a
+  `failed[]` naming each proposal and why it was refused (stale target,
+  validation), instead of reporting `failed: 0` while the same run printed five
+  failures to stderr. `--dry-run` now applies the same stale-target check the
+  real run does, so its prediction stops disagreeing with the outcome. The
+  refusals themselves are correct and unchanged: declining to overwrite a target
+  modified since the proposal was created is the desired behaviour. `akm
+  improve`'s triage pre-pass reports the same count.
+- **Workflow step output is checked against its declared schema (#923).** A step
+  whose returned object does not match its `outputSchema` now says so, naming
+  the step and the specific problem (a missing required field, say). This is a
+  warning: the run continues and its status is unchanged, because a workflow is
+  a guide rather than a contract. Previously an author got neither enforcement
+  nor feedback — the only signal was a notice saying akm was not checking, which
+  is a different fact from whether the output matched.
+
+### Changed
+
+- **A held run lease no longer reports as database corruption (#924).** `akm
+  workflow run` surfaced raw SQLite text — `database is locked`, `database disk
+  image is malformed`, `disk I/O error` — for what was simply another
+  invocation holding the lease. `disk image is malformed` in particular reads as
+  data loss and sent one reporter through integrity checks and a WAL review
+  before finding the real cause. The lease message now appears at default
+  verbosity and carries a dedicated `RUN_LEASE_HELD` code, so a wrapper can tell
+  "retry shortly" from "you passed bad input". Genuine SQLite errors still
+  report as themselves.
+- **`akm lint` stops flagging templated output paths (#927).** A documented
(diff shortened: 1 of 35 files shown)


itlackey:
## Release candidate evidence — `030514b2160c9808c39caef5c81ef9ae1b31ed0e`

### Gated CI — [run 33823559901](https://github.com/itlackey/akm/actions/runs/33823559901), exact SHA, all 7 jobs green

| Job | Result |
| --- | --- |
| Gated / Resolve Candidate | success |
| Gated / Release Candidate Evidence | success |
| Gated / Docker Install | success |
| Gated / Semantic Search | success |
| Gated / Native Scheduler / Linux | success |
| Gated / Native Scheduler / macOS | success |
| Gated / Native Scheduler / Windows | success |

### Local gates on the same SHA

| Gate | Result |
| --- | --- |
| `bunx tsc --noEmit` | clean |
| `bun run lint` | clean (exit 0) |
| `bun run test:unit` | 6559 pass / 5 skip / 0 fail (480/480 files) |
| `bun run test:integration` | 3528 pass / 55 skip / 0 fail (291/291 files) |
| `./tests/release-check.sh --skip-docker` | Release validation passed |

`release-check` was run twice. The first run is **not** cited here: it started before the version-bump commit, so it validated a tree that no longer exists. The result above is from a run against this exact head.

### Scope

Seven issues ship: #917, #921, #922, #923, #924, #926, #927. Four agent branches merged with no conflicts; two gaps that fell between branch boundaries were fixed at merge time — the lease refusal on the `exec` path was still untagged (so `akm workflow run` returned `INVALID_FLAG_VALUE` rather than the new `RUN_LEASE_HELD`), and `improve.ts`'s triage pre-pass omitted the drain `failed` count that #921 adds everywhere else.

#929 and #930 remain open and unimplemented, deferred together for evaluation against the LongMemEval corpus. Investigation since this PR opened found they are the same failure — candidate starvation rather than ranking. Both issues cite the same "1.11 results retrieved" statistic, and because `runFtsQuery` ends in `ORDER BY bm25Score LIMIT ?`, #930's 10:1 title-to-body weighting gates which rows are returned at all, not merely their order.

That investigation also identified a defect underneath both, which neither issue names: `normalizeFtsScores` (`ranking.ts:71-86`) reads `results[0]`/`results[length-1]` as best/worst — assuming a sorted array, which is what forced synthetic bm25 offsets in the #929 attempt — and min-max normalizes over the candidate set, so a document's score depends on what else matched. Measured: three extra weak candidates compress leader separation by 85% and push mid-pack scores toward the 1.0 clamp in `db-search.ts:577`. That is exactly the `graph-boost-ranking` failure, where `oneHopScore` and `twoHopScore` are both exactly `1`. `ranking-contributors.ts:165-186` already documents this pathology and works around it with hard-coded belief-state score ceilings.

Publishing is a separate step (Release workflow with `version: 0.9.13`) and is not part of this PR.

---

Issue #937, opened 2026-09-04
Index long Markdown bodies as lexical fragments without letting one parent monopolize top-K


itlackey:
## Summary

Whole-document FTS rows impose FTS5 length normalization on the entire body. A fact mentioned once in a long document is therefore buried even when the same fact would rank well in a short retrieval unit. The 0.9.14 micro-probe measured a 5.1x penalty for the same body hit at roughly 3,000 tokens versus roughly 80 tokens; changing column weights cannot remove that penalty because FTS5 does not expose `b` or a way to exclude body length from `D`.

Add **lexical fragment retrieval** for Markdown bodies. This is the 0.9.14 fix for long-document burial. It is deliberately separate from semantic/vector chunking (#748) and fragment-aware graph enrichment (#936).

## Dependencies and non-goals

- #933 lands first so changing the candidate population does not rewrite every existing lexical score or saturate downstream boosts.
- #930 is a bounded column-weight experiment, not a substitute for this issue.
- Semantic vectors remain in #748.
- Deterministic and fragment-aware graph work remains in #935/#936.
- Context packing/reassembly is out of scope.

### 1. One shared fragment model

Extract the existing heading parser/section logic and `splitBodyIntoChunks` into one core splitter; do not add a third parser. The result must carry enough identity and provenance to represent:

- content before the first heading;
- normal heading-bounded sections;
- duplicate headings;
- one oversized section split into multiple chunks; and
- headingless transcripts split by paragraph/word fallback.

At minimum each derived fragment needs parent entry identity, deterministic within-revision fragment identity, ordinal, line/offset bounds, optional heading slug, text, and content hash.

`#heading-slug` is not sufficient for headingless or multiply-split sections. Before search returns fragment-qualified refs, define and implement a fragment selector that round-trips through `akm show`: every returned ref must resolve to exactly the indexed fragment text. Do not expose a selector that only works for heading-backed chunks.

### 2. Return top-K distinct parent assets

FTS may score fragment rows, but the user-facing result remains an asset with one winning fragment. A document with many matching fragments must not consume the SQL `LIMIT` before other parents are considered.

- Group/roll up by parent **before** the candidate limit used by ranking.
- Define the parent lexical score as the best matching fragment unless measurements justify a different aggregation; do not sum matches and reward repetition/length.
- Carry the deterministic winning fragment and `lexicalMatch`/`whyMatched` attribution forward.
[...]

Issue #935, opened 2026-09-04
Build the entity graph deterministically — akm already extracts typed relations and throws all of them away


itlackey:
## Summary

The graph is **100% LLM-derived**. Every node, entity and relation in `graph_files` / `graph_file_entities` / `graph_file_relations` comes from `src/llm/graph-extract.ts`. With no configured engine, `akm index` skips extraction entirely (`graph-extraction.ts:754` — "no LLM engine is configured"), `loadGraphBoostContext` returns `null`, and `computeGraphBoost` never runs. Those users have no graph at all.

That would be defensible if the relations genuinely needed a model to find. **They mostly don't.** akm already parses a set of explicit, typed, author-declared edges during indexing — and then drops every one of them on the floor.

## What is already extracted and never used

| signal | where it is parsed | where it goes |
|---|---|---|
| `IndexDocument.links` — resolved `xrefs:` + body markdown links | `llm-wiki-adapter.ts:228` (`resolvePageLinks`), `okf-adapter.ts:159` | assigned at `scan/doc-to-entry.ts:68` and **has no reader anywhere in the codebase** |
| `xrefs:` frontmatter | `metadata.ts:474-479` | concatenated into the FTS `hints` column as bare text (`search-fields.ts:61`) |
| `supersededBy:` / `contradictedBy:` | `metadata.ts:169-172` | belief-state ranking only |
| `derivedFrom:` | `metadata.ts:193` | memory lineage only |
| `refs:` in memory frontmatter | `lint/base-linter.ts:479` (`extractFrontmatterRefs`) | the linter only |
| `sources:` / `evidenceSources:` | `doc-to-entry.ts:128` | provenance display only |

`entry.links` is the sharpest case: it is computed, assigned to the entry, serialized into `document_json`, and **queried by nothing**. It is a finished, deterministic, typed edge list sitting unused in the index.

## Why this matters now

Three separate consequences, and the third is the one that blocks other work.

1. **Users without an LLM engine get nothing.** Graph boost is a documented ranking contributor that a large share of installs can never receive.
2. **The LLM re-derives what the author already declared.** A page with `xrefs: [foo, bar]` and three body links to `baz` has its relationships stated explicitly. Paying a model to guess at them is wasteful and strictly less accurate than reading them.
[...]

Issue #934, opened 2026-09-04
A non-canonical index generation warns once and then fails query-by-query, instead of degrading to the "index not usable" branch callers already have


itlackey:
## Summary

When `index.db` carries a generation this binary does not understand, the read path warns once and then reads it anyway — so every subsequent query fails on its own with a raw SQLite error. Callers already have an "index not usable yet" branch, and it is never taken.

Split out of #928, where this is the akm-side half. The other half of that issue (an old *bundled* binary reading a newer index) is fixed by the openpalm bump and is not in scope here.

## Current behaviour

`warnIfNonCanonicalIndexGeneration` (`src/storage/repositories/index-connection.ts:147`) says exactly what it does:

```ts
warnOnce(
  `index-read-noncanonical:${resolvedPath}`,
  `Index database at ${resolvedPath} does not match this akm's derived schema (stored generation ` +
    `${classification.storedVersion ?? "unknown"}; this binary understands ${CANONICAL_INDEX_DB_VERSION}). ` +
    "Reading it as-is; a query that needs a table or column this generation lacks will fail on its own. " +
    "Run 'akm index' to rebuild it for this binary.",
);
```

Nothing wraps the queries that follow. `searchFts` / `runFtsQuery` (`index-fts-repository.ts`) has no `catch` around its SQL, and its caller at `db-search.ts:612` does not wrap the call either. So a single search produces one explanatory warning followed by N raw failures.

That is what #928 reported from the field:

```
[db] runFtsQuery failed: no such column: e.entry_json
[db] runFtsQuery failed: no such column: e.entry_json
   ... (14 repetitions in one dispatch)
```

with the reporter's actual complaint being that *"the repetition buries real errors in the same stream … ~14 lines of `runFtsQuery` noise around one meaningful line."* One warning ahead of the noise explains its cause; it does not remove it.

## Why this is a gap rather than the intended design

The 0.9.12 guard audit targeted this exact call site. It was a hard `INDEX_SCHEMA_INCOMPATIBLE` refusal on every `openExistingDatabase` / `openReadonlyExistingDatabase` caller, and the finding asked for:

> Degrade: treat a non-canonical generation the way a missing index is treated, so callers take their own "index not usable yet" branch. `improve --dry-run` already models this (`commands/improve/eligibility.ts:53` has an `incompatible` status).

[...]

Issue #933, opened 2026-09-04
normalizeFtsScores makes a document's score depend on what else matched, saturating the ranking ceiling and blocking any recall fix


itlackey:
## Summary

`normalizeFtsScores` (`src/indexer/search/ranking.ts:71-86`) has two structural defects. Together they mean **a document's keyword score depends on which other documents happened to match**, and that any change which widens the candidate pool degrades ranking rather than improving it.

This is the blocker underneath #929 and #930. Neither of those issues names it, and neither can be shipped safely until it is fixed.

```ts
export function normalizeFtsScores(results: DbSearchResult[]): Map<number, { score: number; result: DbSearchResult }> {
  const bestBm25 = results[0]!.bm25Score;
  const worstBm25 = results[results.length - 1]!.bm25Score;
  const range = bestBm25 - worstBm25;

  for (const result of results) {
    const normalized = range !== 0 ? (result.bm25Score - worstBm25) / range : 1.0;
    const ftsScore = 0.3 + normalized * 0.7;
    ftsScoreMap.set(result.id, { score: ftsScore, result });
  }
}
```

## Defect 1 — it assumes the array is sorted by bm25

`bestBm25` and `worstBm25` are read from the first and last elements rather than derived from the array. Nothing in the type or the signature enforces that ordering.

Today `searchFts` returns a single tier, so the assumption holds by accident. The moment more than one tier contributes — which is exactly what #929 does — it breaks, because each tier's bm25 values are on their own scale: a thin match against a strict multi-term AND can score numerically worse than a strong single-term OR hit.

The #929 attempt worked around this by **fabricating score offsets**, nudging each looser tier's rows past the worst score collected so far to keep the array monotonic. That is a patch around this function's assumption, not a fix, and it means the ranker consumes partly synthetic bm25 values.

Fix: derive `min`/`max` from the whole array. This is small and safe on its own, and it lets the synthetic offsets be deleted.

## Defect 2 — min-max normalization makes scores set-dependent

The larger problem. Min-max maps the best hit to `1.0` and the worst to `0.3` **regardless of absolute quality**, so adding a weak candidate changes the score of every other document.

Measured, holding three leaders fixed and adding three genuinely weaker candidates:

| | leaders' ftsScore | leader separation |
| --- | --- | --- |
| 3 candidates | `1.0000`, `0.9300`, `0.3000` | `0.0700` |
| + 3 weaker ones | `1.0000`, `0.9892`, `0.8923` | `0.0108` |

[...]

PR #938, opened 2026-09-04
docs: tighten the 0.9.14 retrieval closeout plan

### LINKED ISSUE #929: searchFts short-circuits the FTS cascade, starving multi-fact queries of candidates

itlackey:
## Summary

- move the 0.9.14 retrieval brief into `docs/plans/` under the path already intended by the issue tracker
- narrow the release to #934, #933, the gated #930 experiment, and lexical fragments in #937
- record #929 / PR #931 as a measured rejection rather than assumed-good preserved work
- defer semantic fragments (#748) and graph expansion (#935/#936) to 0.10.0
- add missing fragment identity, distinct-parent top-K, incremental schema, evaluator pinning, and exact-SHA release gates

## Tracker changes already applied

- created #937 for lexical fragment retrieval
- closed #929 and PR #931 as the preserved negative result
- moved #748/#935/#936 to 0.10.0 with design amendments
- tightened acceptance criteria on #933/#934/#930
- rewrote the 0.9.14 milestone description around four bounded deliverables

## Verification

- `git diff --cached --check`
- reconciled against current `main` code paths and the live GitHub milestone
- verified the pinned `itlackey/akm-eval` probe CLI and exact metrics it grades

Documentation only; no product code changed.

### DIFF
--- a/docs/plans/search-ranking-0.9.14-brief.md
+++ b/docs/plans/search-ranking-0.9.14-brief.md
@@ -1,12 +1,16 @@
 # Search ranking and fragment indexing: brief for 0.9.14
 
-**Status:** investigation complete, no code changed. Written after #929 was
-implemented, measured, and held out of 0.9.13.
-**Scope:** #933, #929, #930, #748, and the two graph gaps it exposed (#935, #936).
+**Status:** critically reviewed against `main` and the live milestone on
+2026-09-04; no product code changed. Written after #929 was implemented,
+measured, and held out of 0.9.13.
+**0.9.14 scope:** #934, #933, #930, and #937. #929 is resolved by rejection
+after its measured zero-recall/large-precision regression. Semantic fragments
+(#748) and graph expansion (#935/#936) move to 0.10.0.
 **Companions:** [`benchmark-tuning-findings.md`](./benchmark-tuning-findings.md)
 is the source of truth for every benchmark figure not measured here.
-**Preserved work:** `issue/0.9.13-search` at `8c61f1f0` — correct in isolation,
-reuse it, do not rewrite it.
+**Preserved negative result:** `issue/0.9.13-search` at `8c61f1f0` and draft
+PR #931. Keep the branch as evidence; do not merge or reuse the union unless a
+new measurement overturns the result below.
 
 ---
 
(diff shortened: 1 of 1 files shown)


itlackey:
Superseded by the 0.9.14 integration PR #939. The reviewed brief commit is the base of release/0.9.14, so its history and review context are preserved there.

PR #941, opened 2026-09-04
fix(search): order clamped relaxed candidates by their pre-clamp score

### LINKED ISSUE #940: Relaxed non-name score ceiling flattens ranking to alphabetical filename order on body-only corpora
## What

`applyRelaxedLexicalScoreCeiling` (`src/indexer/search/ranking.ts`) clamps any relaxed-tier candidate whose **name** holds no query token to a hard `RELAXED_NON_NAME_SCORE_CEILING = 0.65`, and discards the pre-clamp score.

When *no* document name holds a query token — any corpus where `name` is an opaque id — **every** candidate clamps to the identical value. `buildSearchResultComparator` then finds equal scores, equal rounded scores, equal name tiers and equal type boosts, and falls through to its last resort:

```ts
return typeDiff || a.filePath.localeCompare(b.filePath);
```

Alphabetical filename order decides which documents come back.

## Measured

On the LoCoMo retrieval probe (akm-cli 0.9.13, 40 questions, one dialogue turn per document, names like `D1:3`):

- **109 of the returned hits scored exactly 0.65**
- **24 of 40 questions had a top-5 that was entirely tied**

So ~60% of that pack's score was decided by filename order, not relevance.

This is the shape of every conversational or memory corpus — captured memories, imported notes, chat logs, anything whose name is a generated id rather than a title. On those, akm's ranking silently degrades to alphabetical.

It also explains two earlier confusions:

- The probe reference stopped reproducing across runs of the **same binary** (itlackey/akm-eval#16). Nothing about retrieval changed; a different subset of exactly-tied candidates surfaced.
- The #929 cascade patch measured as pure noise. It widened a candidate pool whose scores all collapse to one value, so there was nothing for it to rank.

## Fix

Record the pre-clamp score as `preCeilingScore` — the idiom `applyBeliefStateScoreCeiling` already uses — and have the comparator order on it before reaching the filename compare. Displayed scores are unchanged; only order within a clamped set moves.

LoCoMo probe, 0.9.13 → fixed:

| metric | 0.9.13 | fixed | change |
|---|---|---|---|
| evidenceRecall@5 | 0.564 | 0.692 | +22.7% |
| recall@5 | 0.485417 | 0.591667 | +21.9% |
| precision@5 | 0.227917 | 0.257917 | +13.2% |
| MRR | 0.36625 | 0.4725 | +29.0% |
| nDCG@5 | 0.380024 | 0.492014 | +29.5% |

LongMemEval unchanged on every metric — its names carry query tokens, so the ceiling rarely binds.

## Relationship to #930

#930 asks for FTS5 field-weight work "after stable score normalization". This is that prerequisite: while the ceiling flattens the clamped set, no weight change can be measured, because the ordering that would show the difference is being overwritten.

https://claude.ai/code/session_01K5kmHWR6yvkSMEpFGCMEnt

itlackey:
Closes #940. Unblocks #930.

## Problem

`applyRelaxedLexicalScoreCeiling` clamps relaxed-tier candidates whose name holds no query token to `RELAXED_NON_NAME_SCORE_CEILING = 0.65` and throws the pre-clamp score away. On a corpus where *no* name holds a query token — any conversational or memory corpus, and both benchmark packs — every candidate clamps to the same value and `buildSearchResultComparator` falls through to `a.filePath.localeCompare(b.filePath)`.

Measured on the LoCoMo probe at 0.9.13: **109 of the returned hits scored exactly 0.65, and 24 of 40 questions had a fully tied top-5.** Filename order was deciding ~60% of the pack.

## Change

Two files, 20 lines. The ceiling records `preCeilingScore` (the idiom `applyBeliefStateScoreCeiling` already uses), and the comparator orders on it before reaching the filename compare.

**Displayed scores are unchanged** — only the order within an already-clamped set moves. That is why no existing score assertion shifts.

## Measured effect

LoCoMo retrieval probe, published 0.9.13 → this branch:

| metric | 0.9.13 | this branch | change |
|---|---|---|---|
| evidenceRecall@5 | 0.564 | **0.692** | +22.7% |
| recall@5 | 0.485417 | **0.591667** | +21.9% |
| precision@5 | 0.227917 | **0.257917** | +13.2% |
| MRR | 0.36625 | **0.4725** | +29.0% |
| nDCG@5 | 0.380024 | **0.492014** | +29.5% |

LongMemEval unchanged on every metric (0.95 / 0.676667 / 0.95). Three consecutive probe runs on this branch are identical to every digit.

Recall improves as well as ordering because the clamped pool is reordered *before* truncation to top-K, so better candidates now survive the cut.

## Tests

New red-checked regression test in `tests/scoring-pipeline.test.ts`, alongside the existing "Issue #14: Deterministic sort on tied scores" block. Three documents whose names carry no query token and whose body relevance is the exact inverse of alphabetical order; it fails on `main` and passes here.

Full suite compared against clean `origin/main`:

- `origin/main` (1cd0ee4b): 139 pass, **5 fail**
- this branch: 140 pass, **4 fail**

The failures are a strict subset — all pre-existing (two workflow canonical-ref collisions, one `resolveJudge` freeze, one SSRF DNS case) plus one network flake that happened not to fire on this run. **No new failures.**

https://claude.ai/code/session_01K5kmHWR6yvkSMEpFGCMEnt

### DIFF
--- a/src/indexer/search/db-search.ts
+++ b/src/indexer/search/db-search.ts
@@ -327,6 +327,11 @@ function buildSearchResultComparator(query: string): (a: RankedEntryInput, b: Ra
     if (scoreDiff !== 0) return scoreDiff;
     const rawScoreDiff = stableRankScore(b.score) - stableRankScore(a.score);
     if (rawScoreDiff !== 0) return rawScoreDiff;
+    // Both scores were clamped to the same ceiling. Order by what they scored
+    // BEFORE the clamp rather than falling through to the alphabetical
+    // filePath tie-break, which is not a relevance signal at all.
+    const ceilingDiff = stableRankScore(b.preCeilingScore ?? b.score) - stableRankScore(a.preCeilingScore ?? a.score);
+    if (ceilingDiff !== 0) return ceilingDiff;
     const nameDiff = bNameTier - aNameTier;
     if (nameDiff !== 0) return nameDiff;
     const typeDiff = typeBoostFor(b.entry.type) - typeBoostFor(a.entry.type);
--- a/src/indexer/search/ranking.ts
+++ b/src/indexer/search/ranking.ts
@@ -232,8 +232,21 @@ export function lexicalNameMatchTier(entry: IndexDocument, queryTokens: string[]
  * A relaxed OR query admits intentionally weak candidates. Candidates with no
  * query token in their name remain visible for body-only recall, but cannot
  * saturate at the same displayed score as stronger name-bearing recoveries.
+ *
+ * The pre-clamp score is recorded as `preCeilingScore` (the same idiom
+ * `applyBeliefStateScoreCeiling` uses) so the clamp demotes without also
+ * DESTROYING the relative order of what it demotes. On a corpus whose names
+ * carry no query tokens at all — any conversational or memory corpus, where
+ * `name` is an opaque id — every candidate clamps to the identical ceiling,
+ * every comparison inside the clamped set falls through to the alphabetical
+ * `filePath` tie-break, and the ranking stops measuring relevance entirely.
+ * Measured on the LoCoMo probe before this fix: 109 of 200 returned hits
+ * scored exactly 0.65 and 24 of 40 questions had a fully tied top-5.
  */
 function applyRelaxedLexicalScoreCeiling(item: RankedEntryInput, queryTokens: string[]): void {
   if (item.lexicalMatch !== "relaxed" || lexicalNameMatchTier(item.entry, queryTokens) > 0) return;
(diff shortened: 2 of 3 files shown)


itlackey:
Superseded by the reviewed adaptation in release PR #939 (commit d1b88cb3). The diagnosis and LoCoMo gain remain valid. The release version preserves relaxed pre-ceiling relevance in a separate field so a later belief-state ceiling cannot overwrite it, adds a compound relaxed+archived regression, and updates the display-score comment for #933's monotone public projection. Keeping #940 open until #939's exact-SHA CI and paired akm-eval evidence pass.

